PR #3, opened 2023-03-10
Add acknowledgements page to thank Sponsors etc

robobario:
Lets say thanks to any helpers!


robobario:
Zulip has helped us out with a free OS license, even if we don't use it we should say thanks

SamBarker (approved):
Good idea @robobario 

We should probably also include GitHub for the use of actions and web hosting too.

Issue #6, opened 2023-09-25
Refactor to use Jekyll


robobario:
We want the front website to host a blog and quickstart. Refactoring to Jekyll will make the site blog aware and it's integrated with Github so we can avoid maintaining an action to deploy it.

k-wall:
Using Jekyll seems like a reasonable choice to me.

gracegrimwood:
Fixed by #7. Will close this

PR #7, opened 2023-10-17
Use jekyll for Kroxylicious website

gracegrimwood:
Switches website to use Jekyll for static page generation. This change does not affect the docs at all, which are generated and deployed via the [main Kroxylicious repository](https://github.com/kroxylicious/kroxylicious/blob/main/.github/workflows/asciidoctor-ghpages.yml).

You can try out the site locally yourself by following the instructions in the README. You'll need to have Ruby 2.5.0+ installed, along with any other [minimum requirements for running Jekyll](https://jekyllrb.com/docs/) in order to build and serve the site. Note that the new GitHub workflow takes care of all of this for you when deploying to the web, so these steps are only necessary for running locally.

Note that there is some missing content on some pages (i.e. the default Jekyll blog post and the "to do" placeholders on the quickstarts page), which will follow in seperate PRs once this is merged. I figured this was big enough on its own without adding that to the mix.

Feedback and suggestions are very much welcomed! 🤗


gracegrimwood:
Fixes #6

README.md
```diff
@@ -2,20 +2,47 @@
...
-There is a github action that builds and deploys the HTML/CSS
+There is a GitHub action that builds and deploys the HTML/CSS
 to the `gh-pages` branch on push to `main`.
+
+We don't use the Bootstrap 5 ruby gem here, as it runs on a
+`dart-sass` implementation that's incompatible with the one Jekyll uses.
+
+### Running Locally
+
+To run the site locally, you'll first need to update the [Gemfile](Gemfile) by uncommenting the `jekyll` gem on line 10 and commenting out the `github-pages` gem on line 15.
```
k-wall, line 25: Referencing line numbers in another file is fragile.  They are bound to get forgotten and rot.   Maybe a comment here directing the reader to the Gemfile and comment there telling the developer what to do to run locally.

README.md
```diff
@@ -2,20 +2,47 @@
...
+
+You'll then need to download the Bootstrap 5 Sass sources, which can be done with the following commands:
+
+```bash
+wget "https://github.com/twbs/bootstrap/archive/v5.3.2.zip" # download the full Bootstrap 5 sources (including JS)
```
k-wall, line 30: Let's make this a script.

README.md
```diff
@@ -2,20 +2,47 @@
...
+mkdir ./_sass/bootstrap # make a directory to put the Bootstrap 5 Sass sources into
+mv ./_sass/bootstrap-5.3.2/scss ./_sass/bootstrap/scss # move the sources to your new directory
+rm -r ./_sass/bootstrap-5.3.2 # delete all the other Bootstrap sources (i.e. except the Sass ones, which we moved)
+rm v5.3.2.zip # delete the downloaded zip
+```
+
+Then install the site's prerequisites by running this:
+
+```bash
+bundle install
```
k-wall, line 41: The `bundle install` command failed for me (unprivileged user).

```
Gem::FilePermissionError: You don't have write permissions for the /Library/Ruby/Gems/2.6.0 directory.
```

I don't like letting developer tools update files owned by the OS.

So after some random googling, instead I did:

```
bundle install --path ~/.bundle
```

with the intent of installing stuff into my home account.  It seemed to work.

k-wall:
I don't find the navigation particularly intuitive. How did we decide what goes where?

My eye went straight the navigation in the middle of the page. I was confused as I knew there ought to be a blog option.  It took me a while for my eye to clock the other navigation at the top.  Maybe I'm just being special.
 

Screenshot 2023-10-17 at 16 42 54

README.md
gracegrimwood: Done!

README.md
gracegrimwood: I was a bit dubious about referencing line numbers there myself, but I worried there might bee too many comments already in the Gemfile to be able to easily parse out the ones pointing out which parts to change. I've gone ahead and made the change anyway, and maybe we can look to remove some of the extraneous comments that Jekyll put in there at a later date.

gracegrimwood:
> How did we decide what goes where?

I was trying to recreate the existing [kroxylicious.io](https://kroxylicious.io) home page as best I could in Bootstrap, which is a completely different framework to the one currently used (Bulma). This meant some rearranging of the page:

1. The Kroxylicious vertical logo moved from the right side of the content to the left
2. The "Kroxylicious" header got removed as extraneous content because the word "Kroxylicious" is already in the logo
3. The tagline and link list moved under the logo as they looked out of place above the card in their new stylings
4. The card moved further up the screen as the content above it went elsewhere

image

Other minor changes:

- Background colour changed from `#dce1de` (`$kroxy-light`, a brand colour) to `#ffffff` (`--bs-body-light`, a default Bootstrap colour) to improve contrast and readability
- The word "Kroxylicious" at the start of the card content was italicised to add emphasis and make up for the removal of the title
- Extra spacing was added before the "Why?" header in the card content
- The slack invite link was removed because I couldn't figure out where to put it

Also, as a bit of an aside (which I probably should have put in the original PR comment): I went with [Bootstrap 5](https://getbootstrap.com/) over [Bulma](https://bulma.io/) because Bootstrap has a few more features which made styling the quickstart/blog pages and nav elements a little easier, has built in support for screen readers and other accessibility features, and its a bit more widely used than Bulma which makes finding resources and support a bit easier. There's a comparison of the two on the Bulma site [here](https://bulma.io/alternative-to-bootstrap/) _(it's worth noting to their point about Bulma having "No JavaScript" that the JS component of Bootstrap is completely optional, as is all the JQuery stuff)_.

Hope that explains why it is the way that it is; it's the best I could come up with on my own 😉 . Happy to make changes if people have suggestions on what specifically should be changed (for example, I could perhaps try to make the navbar options more obvious by increasing text size or changing them to a higher contrast colour, if people think that would help).

.github/workflows/jekyll-gh-pages.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Pages
+        uses: actions/configure-pages@v3
+      - name: Download Bootstrap 5 dependencies
+        run: |
+          wget "https://github.com/twbs/bootstrap/archive/v5.3.2.zip" && \
+          unzip v5.3.2.zip -d ./_sass && mkdir ./_sass/bootstrap && \
+          mv ./_sass/bootstrap-5.3.2/scss ./_sass/bootstrap/scss && \
+          rm v5.3.2.zip && rm -r ./_sass/bootstrap-5.3.2
```
SamBarker, line 38: Seems a bit sad to do this on every build but I'm not really convinced about the merits of maintaining our own image (if thats even possible). 

Separately but relatedly do we have to depend on 5.3.2? Is there a way for us to depend on 5.3.latest or even 5.latest ?

_config.yml
```diff
@@ -0,0 +1,52 @@
...
+# https://learnxinyminutes.com/docs/yaml/
+#
+# Site settings
+# These are used to personalize your new site. If you look in the HTML files,
+# you will see them accessed via {{ site.title }}, {{ site.email }}, and so on.
+# You can create any custom variable you would like, and they will be accessible
+# in the templates via {{ site.myvariable }}.
+
+title: Kroxylicious
+description: Network proxy framework for Apache Kafka
```
SamBarker, line 22: Not something that needs fixed now, but we should pick a short description for the project. We have one here and a different one in the [pom](https://github.com/kroxylicious/kroxylicious/blob/40d8ed87f76b82188ef08639fdd71d2f3f7bd0a6/pom.xml#L61) neither of which seem like a great answer for our current state.

acknowledgements.markdown
```diff
@@ -0,0 +1,15 @@
...
+title: Acknowledgements
+permalink: /acknowledgements/
+tagline: "An open-source network proxy framework for Apache Kafka"
+---
+## Acknowledgements
+
+*Thank you for supporting Kroxylicious!*
+
+<img src="{{ '/images/zulip-icon-circle.svg' | absolute_url }}" style="height: 1.5em; vertical-align: middle;">
+<span>&nbsp;<a href="https://zulip.com">https://zulip.com</a></span>
```
SamBarker, line 12: Given we don't actually use zulip should we remove the reference?

@robobario is there anything we should do to shutdown/mothball the zulip instance?

local_setup.sh
```diff
@@ -0,0 +1,6 @@
+wget "https://github.com/twbs/bootstrap/archive/v5.3.2.zip" # download the full Bootstrap 5 sources (including JS)
+unzip v5.3.2.zip -d ./_sass # unpack the sources into the _sass/ directory
+mkdir ./_sass/bootstrap # make a directory to put the Bootstrap 5 Sass sources into
+mv ./_sass/bootstrap-5.3.2/scss ./_sass/bootstrap/scss # move the sources to your new directory
+rm -r ./_sass/bootstrap-5.3.2 # delete all the other Bootstrap sources (i.e. except the Sass ones, which we moved)
+rm v5.3.2.zip # delete the downloaded zip
```
SamBarker, line 6: as @robobario suggested on slack we should setup a docker image for local development but that doesn't need to be included in this PR.

_includes/nav.html
```diff
@@ -0,0 +1,28 @@
...
+          <a class="nav-link" href="https://github.com/kroxylicious/kroxylicious/releases">Download</a>
+        </li>
+        <li class="nav-item me-3">
+          <a class="nav-link" href="https://github.com/kroxylicious">GitHub</a>
+        </li>
+        <li class="nav-item me-3">
+          <a class="nav-link" href="{{ '/blog/' | absolute_url }}">Blog</a>
+        </li>
+      </ul>
+    </div>
```
SamBarker, line 26: I found it quite strange to mix internal and external links like this. 

At a minimum I think we should group `Quickstarts` and `Blog` together and first. 

I'm sure we previously talked about targeting three different personas. 

1. Filter Authors
2. Administrators
3. Managers/Decision makers.

I wonder if we should be providing Nav headings targeting each of those?

SamBarker (approved):
Thanks for picking this up @gracegrimwood I think the new layout looks good. 

It would be really nice to extend this styling to the docs as well, but thats definitely something to be picked up after we have fleshed out the content here.

.github/workflows/jekyll-gh-pages.yml
gracegrimwood: > Separately but relatedly do we have to depend on 5.3.2? Is there a way for us to depend on 5.3.latest or even 5.latest ?

I couldn't find a way to do so with the Sass sources, except maybe for downloading a zip of the latest (and potentially unstable) main branch sources in the repo, instead of the zip of the (stable) v5.3.2 tag sources which we download currently.

Everything is easier when you go via CDN, except you can't get the Sass sources (which have all the functions for inserting your own brand colours) over CDN.

_config.yml
gracegrimwood: Agreed. Everything I've put in here has come from the existing website in some form or another, but it's definitely worth us revisiting this content.

I foresee many content update PRs in our future 😆

SamBarker:
>My eye went straight the navigation in the middle of the page. I was confused as I knew there ought to be a blog option. It took me a while for my eye to clock the other navigation at the top. Maybe I'm just being special.

I did something similar so maybe we should include `quickstarts`and `blog` in the body nav links as well however I think the header links work for consistent navigation across the site (as in they appear on every page which is a fairly standard idiom on the web these days).

We really need to work on/have an elevator pitch for the project. As well as some other content for the home page, a single floating tile looks a bit odd to me.

acknowledgements.markdown
gracegrimwood: This is a holdover from the existing website ([kroxylicious.io/acknowledgements.html](https://kroxylicious.io/acknowledgements.html)). I wasn't sure if it still needed to be there or not, so I kept it in to be safe.

_includes/nav.html
gracegrimwood: > I found it quite strange to mix internal and external links like this.

Yeah, good point. I just wasn't really sure where else to put links to where to get Kroxylicious, so they ended up there. Perhaps these could live on their own page, or could become icons instead of text links (similar to what [Strimzi](https://strimzi.io/) do with their site nav). I agree that `Quickstarts` and `Blog` should be grouped.

> I wonder if we should be providing Nav headings targeting each of those?

They would all lead to the same page (Quickstarts), so I'm not so sure about separating them in the nav like that. The quickstarts page is currently just placeholder content, but it has a tabbed design that allows for the quickstarts for any number of personas to live on the same page (I've got placeholders for filter authors and admins there currently). We could have a page for "Why should you use Kroxylicious?" that targets the manager persona, but I'm not sure calling people (or groups of people) out directly in the nav is what we want to be doing.

I'm not even sure about using the headings "For Kafka Administrators" and "For Filter Developers" in the quickstarts, I think maybe it could be something more to do with how Kroxylicious will be used instead of who will be using it, like "For Running Kroxylicious" and "For Developing Custom Filters" instead (or some less verbose alternative). I don't know, maybe I'm overthinking it.

.github/workflows/jekyll-gh-pages.yml
SamBarker: Thanks for checking.

_includes/nav.html
SamBarker: > Yeah, good point. I just wasn't really sure where else to put links to where to get Kroxylicious, so they ended up there. 

I think having a top right link to github is fairly idomatic these days and we do want to drive traffic to the GitHub repos so I think it qualifies for inclusion in the nav bar. 

>They would all lead to the same page (Quickstarts), so I'm not so sure about separating them in the nav like that

hmm I think what I was meaning was quick starts are just one bit of content for a given persona (and don't exist for all). So I'm questioning if `quckstarts` is the Nav entry or if its really a home for `Developer`, `Administrator` and `Overview` content?

acknowledgements.markdown
robobario: I should go shutdown the instance but still think it's okay to have the thankyou here, they were nice enough to give us a license (with this quid pro quo)

SamBarker: Yeah, the `thank you` is ok, maybe we should make it clear that we don't actually use the service... Just worried about people finding a zulip reference and going hunting for us there...

_includes/project-links.html
```diff
@@ -0,0 +1,15 @@
+<div class="row mx-1">
+  {% capture acknowledgements %}{{ "/acknowledgements/" | absolute_url }}{% endcapture %}
+  {% capture list_items %}
+  {% include list-group-item.html url="https://github.com/kroxylicious" content="GitHub Organisation" icon="github" %}
+  {% include list-group-item.html url="http://kroxylicious.io/kroxylicious" content="Kroxylicious Documentation" icon="file-earmark-code" %}
+  {% include list-group-item.html url="https://github.com/kroxylicious/design" content="Design & Discussion" icon="diagram-3" %}
+  {% include list-group-item.html url="https://github.com/kroxylicious/kroxylicious/blob/main/LICENSE" content="License" icon="file-earmark-text" %}
+  {% include list-group-item.html url="https://kroxylicious.slack.com" content="Community & Development chat via Slack" icon="slack" %}
```
robobario, line 8: need to add the slack invite link somewhere, maybe having the invite link as our slack link would be good enough.

acknowledgements.markdown
tombentley: But then it's weird that we're thanking them for something we don't use any more. I think we should:

1. Shutdown the instance, then
2. Remove the acknowledgement (which could be a follow-on PR).

tombentley:
I also completely missed the blog and quickstart links. +1 to having a single navigation element. We could also highlight the one which corresponds to the current page, which would help. 

Don't feel constrained to do something like what we have currently. For example, while the license is _relevant_ it's probably more important for people to find the Quickstart and the Documentation. We could also group all the developer stuff under a single "Contribute" navigation heading, for example.

acknowledgements.markdown
robobario: fair enough, I've deactivated the organisation

gracegrimwood:
@k-wall @SamBarker @tombentley I've had a go at rearranging things in the nav so they're easier to find. [Demo is here](https://gracegrimwood.github.io/kroxylicious.github.io/), please have a look at let me know what you think.

SamBarker (approved):
I'm 💯 favour of merging as is.

I think the new theme is a big improvement. 

My only remaining thought is as we want to encourage contributions to the project should we make the GitHub links more prominent? Happy for any further shuffling to be a separate PR.

k-wall (approved):
ship it!

gracegrimwood:
Merging this as-is, any further changes will be in seperate PRs.

PR #8, opened 2023-11-07
Update jekyll deps

gracegrimwood:
General cleanup tasks for Jekyll dependencies and plugins:

- Removes unused dependencies from Gemfile (including [jekyll-assets](https://github.com/envygeeks/jekyll-assets), which appears to be abandoned)
- Fixes some inconsistent formatting in the Gemfile
- Swaps out `github-pages` gem for `jekyll`, as the former is meant for legacy GitHub Pages deployments (which we aren't using) while the latter is more up to date. Also updates `README.md` to reflect this change
- Adds the `jekyll-sass-converter` plugin to the list of plugins in `_config.yml`, so it picks up and processes the Bootstrap sources in the `_sass/` directory
- Updates all remaining dependencies and bumps the Bundler version used to install them


gracegrimwood:
Forgot to mention: these changes have been tested and are live at [https://gracegrimwood.github.io/kroxylicious.github.io/](https://gracegrimwood.github.io/kroxylicious.github.io/). Just be aware that the docs link doesn't work because we don't ship the docs in this repo.

README.md
```diff
@@ -22,9 +22,7 @@ We don't use the Bootstrap 5 ruby gem here, as it runs on a
 
 ### Running Locally
 
-To run the site locally, you'll first need to update the [Gemfile](Gemfile) by uncommenting the `jekyll` gem and commenting out the `github-pages` gem.
-
-You'll then need to download the Bootstrap 5 Sass sources by running [local_setup.sh](local_setup.sh).
+To run the site locally, you'll first need to download the Bootstrap 5 Sass sources by running [local_setup.sh](local_setup.sh).
```
tombentley, line 25: This file is empty in the PR at the moment.

Gemfile
```diff
@@ -7,21 +7,15 @@ source "https://rubygems.org"
...
 # Happy Jekylling!
-# gem "jekyll", "~> 4.3.2" # To run locally, uncomment this line.
-# This is the default theme for new Jekyll sites. You may change this to anything you like.
-gem "minima", "~> 2.5"
-# If you want to use GitHub Pages, remove the "gem "jekyll"" above and
+gem "jekyll", "~> 4.3.2"
+# If you want to use legacy GitHub Pages, remove the "gem "jekyll"" above and
 # uncomment the line below. To upgrade, run `bundle update github-pages`.
-gem "github-pages", "~> 228", group: :jekyll_plugins # To run locally, comment out this line.
+# gem "github-pages", "~> 228", group: :jekyll_plugins
```
SamBarker, line 13: presumably this should be restored before merge?

Is there a better way to handle this for local testing?

k-wall: > Swaps out github-pages gem for jekyll, as the former is meant for legacy GitHub Pages deployments (which we aren't using) while the latter is more up to date.

This comment ^^^ makes me suspect we should be removing the reference to the github-pages gem.

README.md
gracegrimwood: [The file isn't empty on the PR branch](https://github.com/gracegrimwood/kroxylicious.github.io/blob/update-jekyll-deps/local_setup.sh), but it looks empty in the PR diff because the only difference is file mode bits (`0644`/`-rw-r--r--` in the main branch, `0755`/`-rwxr-xr-x` in the PR branch). Git sees this as being an "empty file" even though it isn't, because it has trouble representing a diff with only file bit changes and no content difference.

Gemfile
gracegrimwood: Keith is correct, `github-pages` is supposed to be commented out as we actually don't need it if we deploy using GitHub actions. The `github-pages` gem is a fork of Jekyll 1.x, which is very old. We want to be using the much newer `jekyll` gem both locally and on the web.

I left it in just in case we ever did want to use the legacy option, but I don't think that's very likely. I'm happy to ~~kill it with fire~~ remove it altogether 😉

k-wall (approved):
Tested by running it up locally. LGTM
I don't know of a way to confirm that the security issues are all addressed.  Happy to wait  and see what the GitHub scanning thinks.

SamBarker (approved):
LGTM

gracegrimwood:
> I don't know of a way to confirm that the security issues are all addressed.

@k-wall FYI you can test this locally by installing the [bundler-audit](https://github.com/rubysec/bundler-audit) gem and running `bundle audit`. It will generate a list of vulnerabilities from the project's `Gemfile`. 🙂

PR #13, opened 2023-11-24
Quickstarts

gracegrimwood:
Finishing quickstarts. Demo [here](https://gracegrimwood.github.io/kroxylicious.github.io/quickstarts/).


gracegrimwood:
development quickstart is basically copied and pasted from the kroxylicious-sample readme, but I'm trying to work smarter not harder 😉

quickstarts/develop.markdown
```diff
@@ -0,0 +1,98 @@
...
+```
+
+<br />
+
+## Run
+
+Build both `kroxylicious-sample` and `kroxylicious-app` with the `dist` profile as above, then run the following command:
+
+```shell
+KROXYLICIOUS_CLASSPATH="kroxylicious-sample/target/*" kroxylicious-app/target/kroxylicious-app-*-bin/kroxylicious-app-*/bin/kroxylicious-start.sh --config kroxylicious-sample/sample-proxy-config.yml
```
SamBarker, line 52: We should definitely replace this with a shell script. Either generated as part of the dist profile or just copied in. (obviously thats out of scope for this PR)

quickstarts/develop.markdown
```diff
@@ -0,0 +1,98 @@
+Kroxylicious' composable filter chains and pluggable API mean that you can write your own filters and apply your own rules to the Kafka protocol.
+
+In this quickstart you will build a custom filter and use it to modify messages being sent to/consumed from Kafka, learn how filter configuration works, discover how to run Kroxylicious with custom filters, and find a starting point for developing your own custom filters with your own rules and logic.
```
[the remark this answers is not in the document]
gracegrimwood: I think it just needed rephrasing instead, to simplify it somewhat. It was a bit needlessly verbose (and still is a little, IMO, but I'm struggling with how to make it any shorter).

quickstarts/develop.markdown
```diff
@@ -0,0 +1,98 @@
...
+
+<br />
+
+## Getting started
+
+#### Prerequisites
+
+To start developing your own custom filters for Kroxylicious, you will need to install **both** [JDK 17](https://openjdk.org/projects/jdk/17/) and [JDK 21](https://openjdk.org/projects/jdk/21/).
```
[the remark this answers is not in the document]
gracegrimwood: I believe you are correct. I've changed this.

quickstarts/deploy.markdown
```diff
@@ -0,0 +1,25 @@
...
+#### Prerequisites
+
+To get started deploying Kroxylicious, you will need to install a Java Runtime Environment (JRE) with minimum version 17, this does not come included with the Kroxylicious JAR files.
+Some operating systems come with a JRE already installed. You can check what Java version you have installed by running the following command:
+
+```shell
+java -version
+```
+
+If you get an error or the command doesn't return anything, you may not have a JRE installed, or your JRE may not have been correctly added to your `PATH` variable.
```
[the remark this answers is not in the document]
gracegrimwood: I'm not sure I agree, particularly not for deployment. The argument could be made to have recommended tools for developers who want to either build their own filters or contribute directly to Kroxylicious, but in the case of admins wanting to give Kroxylicious a try (which is who I had in mind when writing the deployment quickstart) I think we want to keep the number of things they need to install to get Kroxylicious working to a minimum. SDKman is a great tool, but probably not for that audience.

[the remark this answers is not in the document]
SamBarker: I was more envisaging something along the lines of for installing a JDK we suggest SDKMan...

SamBarker (approved):
Looking good @gracegrimwood

quickstarts/deploy.markdown
```diff
@@ -0,0 +1,54 @@
...
+
+In GitHub, all releases since v0.4.0 have an attached `kroxylicious-app-*-bin.zip` file. Download this, and optionally verify the contents of the package with the attached `kroxylicious-app-*-bin.zip.asc` file.
+
+<br />
+
+## Install and Run
+
+#### Installing Kroxylicious
+
+Extract the downloaded Kroxylicious Zip file into the directory you would like to install Kroxylicious in. Ensure the `kroxylicious-start.sh` and `run-java.sh` files in the `bin/` directory within the extracted folder have at least read, write, and execute (`rwx`) permissions for the owner.
```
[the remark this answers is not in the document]
gracegrimwood: True. I guess I was thinking that a particularly inexperienced user might see "read and execute" and think "oh I'll go remove the write permissions then". But I think you're right, and I can omit the write permissions here.

quickstarts/deploy.markdown
```diff
@@ -0,0 +1,54 @@
+Kroxylicious is a Java application based on [Netty](https://netty.io/), which means it will run anywhere you can run a JVM. (That's a lot of places!)
+To keep things simple, we've created a quick setup guide to help.
```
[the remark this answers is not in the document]
gracegrimwood: Yeah on review what I've written doesn't look right. I'll play with the phrasing and see if I can make it make more sense.

quickstarts/deploy.markdown
```diff
@@ -0,0 +1,54 @@
...
+
+<br />
+
+## Getting started
+
+#### Prerequisites
+
+To get started deploying Kroxylicious, you will need to install a Java Runtime Environment (JRE) with minimum version 17. This does not come included with the Kroxylicious JAR files.
```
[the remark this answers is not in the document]
gracegrimwood: Agreed, this needs rephrasing.

quickstarts/deploy.markdown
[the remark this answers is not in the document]
k-wall: If we direct the user to the tgz rather than the zip we avoid needing to mention anything about the permissions.  AFAIK Windows tooling handles the tar/gz file formats.

quickstarts/develop.markdown
```diff
@@ -0,0 +1,97 @@
...
+
+<br />
+
+#### Get the code
+
+The easiest way to learn how to build custom filters is with our `kroxylicious-sample` module, which contains some basic find-and-replace filters for you to experiment with.
+Begin by downloading the latest `kroxylicious-sample` sources from the [Kroxylicious repository](https://github.com/kroxylicious/kroxylicious).
+
+```shell
+git clone git@github.com:kroxylicious/kroxylicious.git
```
k-wall, line 23: Let's use a https url rather than a ssh. Users may not necessarily have generated an ssh key.

quickstarts/deploy.markdown
```diff
@@ -0,0 +1,54 @@
...
+From the configuration file you can specify how Kroxylicious presents each Kafka broker to the Kafka clients, as well as which filters Kroxylicious should use and any configuration for those filters.
+
+<br />
+
+#### Running Kroxylicious
+
+From within the extracted Kroxylicious folder, run the following command:
+
+```shell
+./bin/kroxylicious-start.sh --config config/example-proxy-config.yml
```
k-wall, line 51: Should the quickstarts include getting the user as far as producing/consuming a message?

quickstarts/deploy.markdown
```diff
@@ -0,0 +1,54 @@
+Kroxylicious is a Java application based on [Netty](https://netty.io/), which means it will run anywhere you can run a JVM. (That's a lot of places!)
+To keep things simple, we've created a quick setup guide to help.
+
+<br />
+
+## Getting started
+
+#### Prerequisites
```
k-wall, line 8: Isn't a Kafka cluster running on the default port also a pre-requirement?

SamBarker: I guess a running cluster is, but the port just needs to be in sync with the Kroxylicious config file.

quickstarts/deploy.markdown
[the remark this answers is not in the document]
SamBarker: OH thats a good point @k-wall I didn't mention using tgz vs zip as I thought it was just my bias showing. We should also double check what permissions we ship the scripts with.

quickstarts/deploy.markdown
gracegrimwood: I had made the assumption that if someone was looking at setting up their own Kroxylicious instance that they would already have a Kafka cluster running (why else would they want to do this?), but maybe it's worth calling it out specifically.

quickstarts/deploy.markdown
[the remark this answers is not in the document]
gracegrimwood: @k-wall Only with 3rd party tools. On Windows, 7zip or similar is required to process tar archives or gzipped files. I wanted to go with the most accessibly cross-platform format, which is (unfortunately) zip.

In saying that though, have we ever tested Kroxylicious on Windows? Maybe it doesn't work and I can ignore that platform

quickstarts/deploy.markdown
SamBarker: I don't think we need to go any further than saying we need a cluster running. Can always point people to our friends Strimzi for help getting a cluster going...

quickstarts/deploy.markdown
gracegrimwood: Yes, but this will be in a subsequent PR as I think this might need some consideration.

quickstarts/deploy.markdown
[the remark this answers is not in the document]
gracegrimwood: I've elected to leave the format as `zip` so we aren't excluding Windows users. We can always change this later if we feel we'd rather use `tar`.

gracegrimwood:
I've pushed up what I think is the final version of this, and the demo is live at [https://gracegrimwood.github.io/kroxylicious.github.io/quickstarts/](https://gracegrimwood.github.io/kroxylicious.github.io/quickstarts/). I'll follow this up with another PR to add produce/consume steps once I've worked out exactly what that should entail.

SamBarker (approved):
LGTM

quickstarts/deploy.markdown
```diff
@@ -0,0 +1,59 @@
+Kroxylicious is a Java application based on [Netty](https://netty.io/), which means it will run anywhere you can run a JVM. (That's a lot of places!)
```
k-wall, line 1: rather than deploy.markdown, can we use `.md` to match the convention we have elsewhere.

quickstarts/deploy.markdown
```diff
@@ -0,0 +1,59 @@
...
+Kroxylicious can be downloaded from the [releases](https://github.com/kroxylicious/kroxylicious/releases) page of the Kroxylicious GitHub repository, or from Maven Central.
+
+In GitHub, all releases since v0.4.0 have an attached `kroxylicious-app-*-bin.zip` file. Download this, and optionally verify the contents of the package with the attached `kroxylicious-app-*-bin.zip.asc` file.
+
+<br />
+
+# Install
+
+Extract the downloaded Kroxylicious Zip file into the directory you would like to install Kroxylicious in.
+Ensure the `kroxylicious-start.sh` and `run-java.sh` files in the `bin/` directory within the extracted folder have at least read and execute (`r-x`) permissions for the owner.
```
k-wall, line 39: Zip has maintained file permissions for a long time, so whilst harmless, I think the suggest to check file permissions is superfluous.

k-wall (approved):
lgtm

quickstarts/deploy.markdown
gracegrimwood: I've been following the convention used by Jekyll, which is to use `.markdown` for pages and site content, and `.md` for things like the README that should not be rendered with the rest of the website. There is functionally no difference between using one or the other and I'm happy to change it if that's what we would prefer, but I personally like the distinction it draws between utility files and actual content.

quickstarts/deploy.markdown
gracegrimwood: There are some (probably old) versions of Zip that don't preserve permissions, and I figure it saves us having to answer this question down the line.

PR #14, opened 2023-11-27
Fixes for GitHub Actions workflow and development work

gracegrimwood:
Update workflow to use jekyll instead of github-pages gem, fix docs link for places not using the kroxylicious domain.

We are currently using a GitHub Action that is pulling in the github-pages gem instead of the Jekyll one, so this PR fixes that by using the Jekyll-recommended workflow (rather than the GitHub-recommended one) which installs and sets everything up manually.

There was also a problem with using absolute_url for the docs link when rendering the site on any domain that wasn't kroxylicious.io, where it would append the path incorrectly (this was only a problem for `/docs` because it's not rendered by Jekyll). Using the relative_url filter for this link should fix this for people who want to test their changes under their own GitHub account, while still retaining functionality on the main site.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ We don't use the Bootstrap 5 ruby gem here, as it runs on a
 
 ### Running Locally
 
-To run the site locally, you'll first need to download the Bootstrap 5 Sass sources by running [local_setup.sh](local_setup.sh).
+To run the site locally, you'll first need to download the Bootstrap 5 Sass sources by running [bootstrap_setup.sh](bootstrap_setup.sh).
 
 Then install the site's prerequisites by running this:
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 ## What?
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -22,7 +22,7 @@ We don't use the Bootstrap 5 ruby gem here, as it runs on a
 
 ### Running Locally
 
-To run the site locally, you'll first need to download the Bootstrap 5 Sass sources by running [local_setup.sh](local_setup.sh).
+To run the site locally, you'll first need to download the Bootstrap 5 Sass sources by running [bootstrap_setup.sh](bootstrap_setup.sh).
```
robobario, line 25: I had some platform issues running the bundle install that were fixed by making sure I was on the latest ruby/rake/bundler versions. Maybe worth adding a statement like "tested with ruby xyz, ..."

bootstrap_setup.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+export BOOTSTRAP_VERSION="v5.3.2"
```
robobario, line 3: could move the `v` prefix out of the variable then use 5.3.2 in the remaining bits of the script that haven't used the version variable

robobario (approved):
LGTM

PR #19, opened 2023-12-20
Fix link to GitHub repo in the footer

ppatierno:


k-wall (commented):
I _think_ the list forming under "supporting" comprises entities and organisation that somehow support kroxylicious existence, with github.con being the first.  As such I think linking to github.com is the right, in the context.

ppatierno:
So I got confused about the purpose of it. I was referring to something similar we have on the https://strimzi.io/ (top right side) where we have links and one of them is using GitHub icon but pointing to Strimzi repo.
I see your reference to GitHub repo could be easily lost in the text a few lines above together with the Slack one.
I think using recognizable icons for visitors is much better ... I got confused as it could happen to any other visitor imho.

SamBarker (approved):
While I agree with @k-wall's description I think its fine to point the icon to our repo anyway. 

I'm leaving the final decision to @gracegrimwood

gracegrimwood (approved):
Yeah I think it's fine to point the icon to the repo. In the context (which is thanking GitHub for hosting the repo and site, I guess) it makes just as much sense as linking to the GitHub homepage IMHO.

Thanks for pointing this out Paolo! 🙂

k-wall (approved):

PR #29, opened 2024-01-25
Fix bootstrap colour palette and import order

gracegrimwood:
The primary, secondary, light, and dark colour palette overrides in _scss/kroxylicious.scss were not applying correctly due to an issue with import ordering. This fixes that (and some other nonsense variables we had floating around), so we're actually using the Kroxylicious brand colours across the site.

### DIFF
--- a/_sass/kroxylicious.scss
+++ b/_sass/kroxylicious.scss
@@ -1,8 +1,6 @@
 @charset "utf-8";
 // Required
 @import "./bootstrap/scss/functions";
-@import "./bootstrap/scss/variables";
-@import "./bootstrap/scss/variables-dark";
 
 @import url('https://fonts.googleapis.com/css?family=Josefin+Sans:300,400,700&display=swap');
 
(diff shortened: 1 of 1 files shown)


_sass/kroxylicious.scss
```diff
@@ -49,7 +41,7 @@ $light: $kroxy-light;
   padding: 0.75rem 0;
   background-color: transparent;
   background: linear-gradient(to bottom, rgba($kroxy-dark, 1), rgba($kroxy-dark, 0.95));
-  box-shadow: 0 0.5rem 1rem $kroxy-shadow-dark,inset 0 -1px 0 $kroxy-shadow-dark;
+  box-shadow: 0 0.5rem 1rem rgba($kroxy-dark, 0.15),inset 0 -1px 0 rgba($kroxy-light, 0.15);
```
SamBarker, line 44: Is there a reason you've moved from pre computing `rgba($kroxy-dark, 0.15)` as `$kroxy-shadow-dark` to calling it inline?

SamBarker (approved):
From my limited level of clue LGTM

_sass/kroxylicious.scss
gracegrimwood: This variable is just a transparency of `$kroxy-dark`, and it's only used in one place. We compute other transparencies of this variable elsewhere already (for example the line above this one), so given this scss file is compiled at build time by jekyll-sass-converter into regular CSS anyway, it felt a little redundant to keep it around.

PR #35, opened 2024-01-29
Re-work Hero band and change global font

insectengine:
1) Added hero band with new text as per original PR 
2) Added "row" to layout to remove unwanted global side padding so the page renders at true full width. 
3) change from Josefin to Questrail font.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -12,10 +12,8 @@
 </head>
 <body class="d-flex flex-column vh-100" data-bs-theme="light">
 {% include nav.html %}
-<div class="container-fluid" data-bs-theme="light">
-    <div class="row">
+<div class="container-fluid px-0" data-bs-theme="light">
     {{ content }}
-    </div>
 </div>
 {% include footer.html %}
 </body>
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -13,7 +13,9 @@
 <body class="d-flex flex-column vh-100" data-bs-theme="light">
 {% include nav.html %}
 <div class="container-fluid" data-bs-theme="light">
+    <div class="row">
     {{ content }}
+    </div>
 </div>
 {% include footer.html %}
 </body>
--- a/_layouts/index.html
+++ b/_layouts/index.html
@@ -1,8 +1,31 @@
 ---
 layout: default
 ---
+
+<!-- Hero -->
+
+<div class="krx-hero">
+    <div class="container col-xxl-8 px-4 py-5">
+    <div class="row flex-lg-row-reverse align-items-center g-5 py-5">
+      <div class="col-10 col-sm-8 col-lg-6">
+        <img src="{{site.baseurl}}/images/kroxylicious_logo_rgb_no_text.svg" class="d-block mx-lg-auto img-fluid" alt="Kroxylicious Mascot" width="700" height="500" loading="lazy">
+      </div>
+      <div class="col-lg-6">
+        <h1 class="fw-bold text-body-emphasis lh-1 mb-3">Kroxylicious transparently inject behaviour into your Apache Kafka<sup>&#174;</sup> cluster</h1>
+        <p class="lead">Solutions to cross-cutting concerns in areas including security, policy-enforcement, audit and much more.</p>
+        <div class="d-grid gap-2 d-md-flex justify-content-md-start">
+          <a href="{{site.baseurl}}/quickstarts/" alt="Get started with Kroxylicious" type="button" class="btn btn-light btn-lg px-4 gap-3">Get Started</a>
+          <a href="{{site.baseurl}}/kroxylicious/" alt="Read the documentation" type="button" class="btn btn-light btn-lg px-4">Read the Docs</a>
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
+
+<!-- Hero -->
+
 <div class="row justify-content-center">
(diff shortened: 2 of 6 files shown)


insectengine:
@k-wall  - I ran into all sorts of trouble updating/rebasing the other PR so I closed it and resubmitted here.

_layouts/index.html
```diff
@@ -1,8 +1,31 @@
...
+<!-- Hero -->
+
+<div class="krx-hero">
+    <div class="container col-xxl-8 px-4 py-5">
+    <div class="row flex-lg-row-reverse align-items-center g-5 py-5">
+      <div class="col-10 col-sm-8 col-lg-6">
+        <img src="{{site.baseurl}}/images/kroxylicious_logo_rgb_no_text.svg" class="d-block mx-lg-auto img-fluid" alt="Kroxylicious Mascot" width="700" height="500" loading="lazy">
+      </div>
+      <div class="col-lg-6">
+        <h1 class="fw-bold text-body-emphasis lh-1 mb-3">Kroxylicious transparently inject behaviour into your Apache Kafka<sup>&#174;</sup> cluster</h1>
```
SamBarker, line 14:
```suggestion
        <h1 class="fw-bold text-body-emphasis lh-1 mb-3">Kroxylicious transparently injects behaviour into your Apache Kafka<sup>&#174;</sup> cluster</h1>
```
or
```suggestion
        <h1 class="fw-bold text-body-emphasis lh-1 mb-3">Kroxylicious transparently injecting behaviour into your Apache Kafka<sup>&#174;</sup> cluster</h1>
```

_sass/kroxylicious.scss
```diff
@@ -2,10 +2,10 @@
 // Required
 @import "./bootstrap/scss/functions";
 
-@import url('https://fonts.googleapis.com/css?family=Josefin+Sans:300,400,700&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Questrial&display=swap');
 
 $font-family-sans-serif:
-        "Josefin Sans", system-ui, -apple-system, "Segoe UI", Roboto, "Helvetica Neue",
+        "Questrial", system-ui, -apple-system, "Segoe UI", Roboto, "Helvetica Neue",
```
SamBarker, line 8: Should `system-ui` and `-apple-system` be quoted? or are they magic SCSS things?

SamBarker (approved):
thanks @insectengine 

It looks really good to me.

Attaching a copy of the rendered site for others to view.

Screenshot 2024-01-30 at 10 49 34

_layouts/default.html
```diff
@@ -13,7 +13,9 @@
 <body class="d-flex flex-column vh-100" data-bs-theme="light">
 {% include nav.html %}
 <div class="container-fluid" data-bs-theme="light">
+    <div class="row">
```
gracegrimwood, line 16: This change breaks the quickstart and blog posts layouts, as the gutters in the `card` are off which causes the `card-header` background to render incorrectly. Could this `div` not move to the `index.html` layout?

Alternately, adding `px-0` to the parent container and omitting this `row` entirely seems to solve the issue of excess x-padding on the hero.

_sass/kroxylicious.scss
gracegrimwood: They're magical CSS properties, not to worry. Google says it's fine 😉

_sass/kroxylicious.scss
```diff
@@ -117,4 +117,16 @@ $navbar-dark-disabled-color: rgba($white, 0.45);
...
+}
+
+// Hero Styles
+
+.krx-hero {
+  background: rgb(73,160,120);
+  background: linear-gradient(111deg, rgba(73,160,120,1) 0%, rgba(44,96,72,1) 30%, rgba(22,48,36,1) 100%);
+
+  h1, p {
+  color: $white !important;
```
gracegrimwood, line 129: I don't have a problem with doing it this way, but worth noting the property `data-bs-theme="dark"` would achieve the same effect on all child text.

gracegrimwood (commented):
A few comments on some layout and CSS bits. I'd really like us to address the broken layouts before this is merged.

_sass/kroxylicious.scss
SamBarker: /me moves on quietly

_layouts/index.html
k-wall: Either wording is good for me.

k-wall:
@SamBarker @gracegrimwood Do you think the hero's buttons would be better to direct the reader to the Oveview/Use-Cases.  I think this is going to be a more likely first port of call.

k-wall (approved):

_layouts/default.html
insectengine: Thanks for double checking this. I have limited knowledge of working with bootstrap. Thanks for the fix!

_sass/kroxylicious.scss
gracegrimwood: Seems I was wrong because the `<p>` text in the hero isn't inheriting the colour for some reason. I'm happy to leave this as it is.

_layouts/default.html
gracegrimwood: I've submitted a [PR](https://github.com/insectengine/kroxylicious.github.io/pull/1) on your branch to fix this.

k-wall (approved):
lgtm

PR #36, opened 2024-02-01
Fix #31: Enable anchor links for kroxylicious.io

### LINKED ISSUE #31: Anchor links for kroxylicious.io
It'd be nice if content headings on the kroxylicious.io had copy/pastable anchor links to that users could easily refer to content from other places.

image

k-wall:
It looks like this:

image


css/code.css
```diff
@@ -148,3 +148,23 @@
...
 }
+
+h2>a.anchor, h3>a.anchor, h4>a.anchor, h5>a.anchor, h6>a.anchor {
+  display: block;
+  margin-left: -1.5ex;
+  position: absolute;
+  text-decoration: none !important;
+  visibility: hidden;
+  z-index: 2;
+  transition: visibility 1s;
```
k-wall, line 159: I did this to stop the visibility transition being applied before the user's pointer could reach the chain symbol.

_use_cases/encryption-at-rest.markdown
```diff
@@ -2,7 +2,9 @@
 name: Encryption At Rest
 ---
 
-## Why
+### Encryption At Rest
+
+#### Why
```
k-wall, line 7: I adjusted the nesting level to be in keeping with the overview, which i think looks nicer.

css/code.css
```diff
@@ -148,3 +148,23 @@
 .highlight .w {
   color: #bbbbbb;
 }
+
+h2>a.anchor, h3>a.anchor, h4>a.anchor, h5>a.anchor, h6>a.anchor {
```
k-wall, line 152: The CSS is based on that used by the Openshift documentation.

gracegrimwood (commented):
I've submitted [a PR](https://github.com/k-wall/kroxylicious.github.io/pull/2) to adjust a few things, the details are in the PR description :)

k-wall:
@gracegrimwood thank you.

One issue, can we make the anchoring relative to the black title area?  In this case I would expect the "What is Kroxylicious" to be displayed immediately beneath the black title area.

image

gracegrimwood:
Took some digging because the usual solution (having the anchor link to an invisible div offset vertically up the page from the heading) isn't available to us because of the way we're injecting these anchors into the page. Turns out we can do it with the `scroll-margin-top` property instead, so I've submitted a PR to fix that [here](https://github.com/k-wall/kroxylicious.github.io/pull/3).

PR #38, opened 2024-02-09
Quickstarts Part 2

gracegrimwood:
- Add _Use_ section to deploy quickstart
- Add new `bs_alert.html` ([Bootstrap Alert component](https://getbootstrap.com/docs/5.3/components/alerts/)) include (used within deploy quickstart at the moment, but is reusable enough to put anywhere)
- Convert quickstarts to collection so we aren't publishing the raw markdown files anymore ([example](https://kroxylicious.io/quickstarts/deploy.markdown))
- Add front matter to quickstart files so we can do tabbed content
- Fix `markdownify` filter behaviour in quickstarts


_quickstarts/deploy.markdown
```diff
@@ -0,0 +1,114 @@
...
+./bin/kroxylicious-start.sh --config config/example-proxy-config.yml
+```
+
+To use your own configuration file instead of the example, just replace the file path after `--config`.
+
+<br />
+
+# Use
+
+To use your Kroxylicious proxy, your client(s) will need to point to the proxy (using the configured Port-Per-Broker or SNI address) rather than directly at the Apache Kafka® cluster.
```
k-wall, line 83: SNI can't work without TLS on Kroxy's downstream side.  The feature relies on the TLS client hello. So either the quick start would need to go into setting up TLS and provide example config, we just use port-per-broker for the quickstart.  I'd advocate the latter.

_quickstarts/deploy.markdown
```diff
@@ -0,0 +1,114 @@
...
+
+{% capture jre_note %}
+If you get an error or the command doesn't return anything, you may not have a JRE installed, or your JRE may not have been correctly added to your system's `PATH` variable.
+{% endcapture %}
+
+{% include bs-alert.html type="primary" icon="info-circle-fill" content=jre_note %}
+
+#### Apache Kafka®
+
+You will also need a running Apache Kafka® cluster to point Kroxylicious at. The Strimzi project has an [easy-to-follow quickstart](https://strimzi.io/quickstarts/) for setting up a cluster on your local machine using Minikube, or alternatively you can follow the official Apache Kafka® [quickstart](https://kafka.apache.org/documentation/#quickstart) to set up a local bare metal cluster.
```
k-wall, line 32: I had problems with the rendering of the registered sign.  I had to use the entity ref.

k-wall: how about?
```suggestion
You will also need a running Apache Kafka® cluster to be proxied by Kroxylicious. The Strimzi project has an [easy-to-follow quickstart](https://strimzi.io/quickstarts/) for setting up a cluster on your local machine using Minikube, or alternatively you can follow the official Apache Kafka® [quickstart](https://kafka.apache.org/documentation/#quickstart) to set up a local bare metal cluster.
```

k-wall: The strimzi quickstart gives you a kubernetes cluster within Kubenetes.  You need to do extra stuff to expose the cluster to the host network (nodeports, advertised listeners etc), otherwise the Kroxylicious instance running locally won't be able to connect to it.

_quickstarts/deploy.markdown
gracegrimwood: Agreed, it would be easier to point people at Port-Per-Broker initially.

_quickstarts/deploy.markdown
gracegrimwood: > I had problems with the rendering of the registered sign. I had to use the entity ref.

Fixed!

> how about?

I ended up going with "You will also need a running Apache Kafka&#174; cluster for Kroxylicious to proxy." as I thought it read better in the context of the rest of the quickstart.

> The strimzi quickstart gives you a kubernetes cluster within Kubenetes. You need to do extra stuff to expose the cluster to the host network (nodeports, advertised listeners etc), otherwise the Kroxylicious instance running locally won't be able to connect to it.

I had forgotten about that, good point. In the interest of keeping things simple I've just removed the references to the Strimzi quickstart entirely. Maybe we can have a blog post at some point in the future about setting up Kroxylicious with a Strimzi cluster.

_quickstarts/deploy.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+To use your own configuration file instead of the example, just replace the file path after `--config`.
+
+<br />
+
+# Use
+
+To use your Kroxylicious proxy, your client(s) will need to point to the proxy (using the configured address) rather than directly at the Apache Kafka&#174; cluster.
+
+Here's how you would use Kroxylicious with the console producer and console consumer:
```
SamBarker, line 91:
```suggestion
Here's how you would use Kroxylicious with the command line clients from the Apache Kafka&#174; distribution:
```

It sounds a bit clumsy, but it hints at where to get the tools from and covers the fact the first script used isn't actually the console producer or consumer.

_quickstarts/deploy.markdown
```diff
@@ -0,0 +1,119 @@
...
+
+# create a topic "my_topic" via Kroxylicious
+kaf -b $KROXYLICIOUS_BOOTSTRAP topic create my_topic
+
+# produce the string "hello world" to your topic via Kroxylicious
+echo "hello world" | kaf -b $KROXYLICIOUS_BOOTSTRAP produce my_topic
+
+# Consume your string from your topic via Kroxylicious
+kaf -b $KROXYLICIOUS_BOOTSTRAP consume my_topic
+```
```
SamBarker, line 119: I feel like the titles get a bit lost. What would you think of having three separate examples?

e.g 
````suggestion

{% capture kroxylicious_bootstrap %}
In each command below, substitute $KROXYLICIOUS_BOOTSTRAP for the bootstrap address of your Kroxylicious instance.
{% endcapture %}

{% include bs-alert.html type="primary" icon="info-circle-fill" content=kroxylcious_bootstrap %}
## create a topic "my_topic" via Kroxylicious

```shell
kaf -b $KROXYLICIOUS_BOOTSTRAP topic create my_topic
```

## produce the string "hello world" to your topic via Kroxylicious
```shell
echo "hello world" | kaf -b $KROXYLICIOUS_BOOTSTRAP produce my_topic
```
## Consume your string from your topic via Kroxylicious
```shell
kaf -b $KROXYLICIOUS_BOOTSTRAP consume my_topic
```
````

It would be cool to toggle between kaf and the command line clients instead of duplicating but that feels like a step to far, for this PR at least.

**Review by SamBarker — APPROVED** [review]

File: _quickstarts/deploy.markdown (hunk 3 above)
Comment (gracegrimwood) [reply _quickstarts/deploy.markdown:91]: We're already pointing users to the Apache Kafka quickstart at the start of this doc, which involves using these clients. If we still feel the need to point this out here specifically, I might instead use the phrasing
```suggestion
Here's how you would use Kroxylicious with the Apache Kafka&#174; command line clients:
```

_quickstarts/deploy.markdown
gracegrimwood: > I feel like the titles get a bit lost. What would you think of having three separate examples?

They weren't meant to be titles, so much as steps. Perhaps I could number each step?
I'm reluctant to over-use the alert component, lest the page look too busy. I can put that instruction as a regular paragraph rather than monospace.

> It would be cool to toggle between kaf and the command line clients instead of duplicating but that feels like a step to far, for this PR at least.

I'm more than happy if someone finds a way to nest tabbed components in a way that doesn't break pages, but I think actually implementing something like that would be a lot of work. It's not something I'm planning on looking into right now.

_quickstarts/deploy.markdown
SamBarker: I think it needs a little more context here, after all people don't read, they skim at best. Your version certainly reads better than mine.

_quickstarts/deploy.markdown
SamBarker: >I'm reluctant to over-use the alert component, lest the page look too busy. I can put that instruction as a regular paragraph rather than monospace.

I don't have a rendered version handy just now but I didn't think that was any more or less alert worth than the others (happy to be wrong), but I don't see why it wouldn't be a simple note as you suggest. I certainly wasn't suggesting it should alert for each client! 

>They weren't meant to be titles, so much as steps. Perhaps I could number each step?

No, I don't think they need numbering. I know I mentioned the titles but actually thats not really what I was focused on, I want to give more focus to the individual commands themselves. I don't quite know what why to land so I'm going to brain dump a few thoughts:

- A single block (as the PR) works well as a facsimile of a single terminal session which is a nice property but it blurs the commands together (at least for me). Breaking them out like I suggest makes the commands stand out and their explanations more obvious but the overall experience is less cohesive. 
- I guess the other option would be to put a comment after each command with the number of a footnote and put the comments there. 
- Maybe in the broken out version we should include sample output so people see how the output looks?
- Who are we writing this for and what do we want to show to them? 
- - Is it something easy to copy and paste? Then I'd favour breaking it up. Is it just to show how simple it is to get going? then a single block works better...

gracegrimwood: I've had a go at reworking that section (along with some other bits) and I've managed to add a tabbed card that nests inside the page's larger tabbed card, so now we can switch between clients for the examples. I've made the steps both titles _and_ numbered (compromise! 😄), so at least it's clear to those very new users that they do need to go in that specific order.

> Breaking them out like I suggest makes the commands stand out and their explanations more obvious but the overall experience is less cohesive.

Agreed that it makes them more understandable and obvious, and once I put them inside the tabbed card it made it look cohesive again IMO.

> I guess the other option would be to put a comment after each command with the number of a footnote and put the comments there.

I didn't do this because the idea of having to scroll down to read a footnote and then scroll back up to read the command again sounded a bit painful. I think what I've come up with is a decent compromise.

> Maybe in the broken out version we should include sample output so people see how the output looks?

I haven't added sample output only because the syntax highlighter doesn't know the difference between what is command and what is output, so it highlights them just the same and it looks very confusing. If we find something down the line that's a little smarter than Rouge and knows how to tell the difference, we can always switch them out.

> Who are we writing this for and what do we want to show to them?

The brief I gave myself for the quickstarts was "as broad an audience as possible while still being useful". I'm thinking people, probably in the general categories of admin or developer, who either already know Kafka but don't know Kroxylicious, or who don't know Kafka _or_ Kroxylicious. The only things I'm assuming is that the audience has a base level of technical understanding (i.e. they know what a terminal is and how to find one) and that they know how to read 😉 

> Is it something easy to copy and paste? Then I'd favour breaking it up.

I think having the commands available to copy-paste is ideal, and the changes I've made do make it easier to see the difference between what's a comment and what's a command... But I do want to note that on a functional level there's no difference between having the instructions be part of the code block or seperate. The reader still has to manually select the text and copy-paste it themselves, because we don't have those quick copy-to-clipboard buttons (like GitHub has, for example) on the site's code blocks.

Anyway, please let me know what you think of the latest changes! 😄

SamBarker:
Thanks for all the work @gracegrimwood its looking awesome. 

+1 from me. Lets get this merged.

SamBarker (approved):
Thanks for all the work @gracegrimwood its looking awesome.

+1 from me. Lets get this merged.

PR #39, opened 2024-02-19
Update tag line

SamBarker:
New tag line!

Screenshot 2024-02-19 at 14 42 45

I added Topic encryption to the suggested subtitle as that is the one thing we have today and we are rightly proud of it.

### DIFF
--- a/_layouts/index.html
+++ b/_layouts/index.html
@@ -12,7 +12,7 @@ layout: default
       </div>
       <div class="col-lg-6">
         <h1 class="fw-bold text-body-emphasis lh-1 mb-3">Kroxylicious, the snappy open source proxy for Apache Kafka<sup>&#174;</sup></h1>
-        <p class="lead">Topic encryption, policy-enforcement, multitency, audit and much more.</p>
+        <p class="lead">Topic encryption, policy-enforcement, multi-tenancy, audit and much more.</p>
         <div class="d-grid gap-2 d-md-flex justify-content-md-start">
           <a href="{{site.baseurl}}/quickstarts/" alt="Get started with Kroxylicious" type="button" class="btn btn-light btn-lg px-4 gap-3">Get Started</a>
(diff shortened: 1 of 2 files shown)


_layouts/index.html
```diff
@@ -11,8 +11,8 @@
         <img src="{{site.baseurl}}/images/kroxylicious_logo_rgb_no_text.svg" class="d-block mx-lg-auto img-fluid" alt="Kroxylicious Mascot" width="700" height="500" loading="lazy">
       </div>
       <div class="col-lg-6">
-        <h1 class="fw-bold text-body-emphasis lh-1 mb-3">Kroxylicious transparently inject behaviour into your Apache Kafka<sup>&#174;</sup> cluster</h1>
-        <p class="lead">Solutions to cross-cutting concerns in areas including security, policy-enforcement, audit and much more.</p>
+        <h1 class="fw-bold text-body-emphasis lh-1 mb-3">Kroxylicious, the snappy open source proxy for Apache Kafka<sup>&#174;</sup></h1>
+        <p class="lead">Topic encryption, policy-enforcement, multitency, audit and much more.</p>
```
k-wall, line 15: typo + hyphenation 
```suggestion
        <p class="lead">Topic encryption, policy-enforcement, multi-tenancy, audit and much more.</p>
```

k-wall:
I think the buttons should take us to the overview and use-cases, rather than getting started and docs.  Those are the resources that a new arrival is likely to want first.

SamBarker:
> I think the buttons should take us to the overview and use-cases, rather than getting started and docs. Those are the resources that a new arrival is likely to want first.

I'd contest it should be `use-cases` or `overview` and `quick starts`. I lean towards `use-cases` rather than `overview`.  
As you say those are the important resources for people wanting context but we want to get people up and running with as little fuss as possible. I am starting wonder if we should just move overview below the fold on the homepage? 

The docs are good landing point for people wanting details.

I actually think it renders nicely with three buttons so maybe cover each base

Screenshot 2024-02-22 at 14 37 36

SamBarker:
I've opened https://github.com/kroxylicious/kroxylicious.github.io/pull/40 to take the button conversation further.

PR #42, opened 2024-03-06
Rerendered the encryption-at-rest_solution diagram to include the envelope->record rename.

k-wall:
(also fixed up a couple of white backgrounds (within envelope + gear icons) that showed up when view in dark mode)


robobario (approved):
LGTM

Issue #44, opened 2024-03-20
Render Kroxylicious AsciiDoc on the website with Jekyll


gracegrimwood:
It would simplify things immensely if the Kroxylicious AsciiDoc files were rendered via this repository and its workflows, using Jekyll (i.e. with the [`jekyll-asciidoc`](https://github.com/asciidoctor/jekyll-asciidoc) plugin) rather than being rendered separately via the workflow in the Kroxylicious repository. However, we don't want to move the AsciiDoc files out of the Kroxylicious repository because it is easier to keep these docs up to date when they live closer to the code.

Rendering the AsciiDoc via this repository would mean that styles can be applied uniformly across the site (rather than the very different styles currently used) and would make linking back to the main site from the docs page much easier.

This rendering could easily be added as a step in the build job before [the "Build with Jekyll" step](https://github.com/kroxylicious/kroxylicious.github.io/blob/main/.github/workflows/jekyll-gh-pages.yml#L44), the only question is how to get the files from the Kroxylicious repository to this one so they can be rendered. There are two ways I can think of, each with pros and cons:

1. Push AsciiDoc files into this repository from the Kroxylicious repository via GitHub Actions ([this is how Strimzi does it](https://github.com/strimzi/strimzi-kafka-operator/blob/main/.azure/scripts/docu-push-to-website.sh)), i.e. after every push to main
   - ✅ The site would update every single time a change is made, so nothing is missed
   - ❌ Pushes may happen too frequently (i.e. it might try to push even if no AsciiDoc changes have been made)
   - ❌ We would end up storing the same docs files in two different locations (this repo and the Kroxylicious repo)
   - ❌ We would have to give a bot the ability to push to this repository
2. Pull AsciiDoc into the GitHub Actions image at build time
   - ✅ No need to store the same docs in two different repos
   - ✅ No extra permissions for workflows or bots required, we can just clone the Kroxylicious repo
   - ✅ Don't necessarily need to modify the workflow for the main Kroxylicious repo, so there's no risk of breaking something important (like the release pipeline)
   - ✅ We could use [the `repository_dispatch` trigger](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch) to fire off the website build pipeline after every merge to main in the main Kroxylicious repository, which would keep the website up to date whenever a change is made to the AsciiDoc.
[...]

PR #47, opened 2024-04-21
Render Kroxylicious AsciiDoc on the website with Jekyll

gracegrimwood:
Adds rendering of the docs using `jekyll-asciidoc` (#44).

Also modifies site-wide code highlighting styles to better align with the highlighting used by the development docs (it has better contrast and readability than what we were using before).

### Some notes:

- Code highlighting doesn't work on the docs unless the source highlighter is changed to rouge in the AsciiDoc files (in `index.adoc` and `custom-filters.adoc`), and Jekyll will produce a non-fatal warning on build if this is not done. Alternatively we could install `pygments` specifically for the AsciiDoc pages, but I thought it was preferential to not have two code highlighters installed (Jekyll uses `rouge`).
- If we want to add a version number to the page to indicate which version's docs we're rendering, we would change that in the AsciiDoc too (i.e. by changing the title in `index.adoc`).
- The nav now directs to this new docs page, which then contains a link to the development docs. We could probably make that nav tile into a dropdown with both links, but I wasn't sure I wanted to tackle the CSS for that in this PR.
- The docs files in this PR were sourced from the `v0.5.1` tag in the Kroxylicious repository.


gracegrimwood:
There are issues with the workflow rendering the a2s images. Marking as draft until I fix that.

gracegrimwood:
Fixed workflow so AsciiToSVG images now render correctly. I have had to add Go to the workflow to do this.

SamBarker:
I think that looks a lot better when I check locally. 

Excuse the huge screenshot but I think its helpful to provide a side by side comparison.
Screenshot 2024-04-22 at 12 03 44 PM

> Code highlighting doesn't work on the docs unless the source highlighter is changed to rouge in the AsciiDoc files (in index.adoc and custom-filters.adoc), and Jekyll will produce a non-fatal warning on build if this is not done. Alternatively we could install pygments specifically for the AsciiDoc pages, but I thought it was preferential to not have two code highlighters installed (Jekyll uses rouge).

Lets update the adoc to use `rouge`. One formatter to rule them all....

> If we want to add a version number to the page to indicate which version's docs we're rendering, we would change that in the AsciiDoc too (i.e. by changing the title in index.adoc).

Also worth doing, but I was expecting the website to maintain docs for each version not just a single copy of the current release docs. So `/docs/0.5.0` & `/docs/0.3.0` would work was my expectation. That said I'm fine with this PR just fixing the rendering and sorting versioning in a separate PR.

Gemfile.lock
```diff
@@ -3,17 +3,23 @@ GEM
...
     em-websocket (0.5.3)
       eventmachine (>= 0.12.9)
       http_parser.rb (~> 0)
     eventmachine (1.2.7)
-    eventmachine (1.2.7-java)
-    eventmachine (1.2.7-x64-mingw32)
     ffi (1.16.3)
     forwardable-extended (2.6.0)
-    google-protobuf (4.26.1)
+    google-protobuf (4.26.1-arm64-darwin)
```
SamBarker, line 22: Hmmm doesn't seem right that we are referencing arch specific things here, especially not darwin. Do we really need `Gemfile.lock` under source control?

SamBarker (commented):
I think if we can move the docs under a version specifier then I'm +1 on this change.

SamBarker:
Local builds also show a couple of warnings 
```shell
asciidoctor: WARNING: available-filters/record-encryption/record-encryption.adoc: line 158: section title out of sequence: expected level 4, got level 5
asciidoctor: WARNING: available-filters/record-encryption/record-encryption.adoc: line 158: section title out of sequence: expected level 4, got level 5
```
I presume those are also issues in the runtime source tree?

Gemfile.lock
gracegrimwood: Do we need it? No, strictly speaking we don't. Should we keep it there anyway? Well, according to the bottom of [this page](https://bundler.io/guides/faq.html) in the official Bundler docs:

> Q: Should I commit my `Gemfile.lock` when writing a gem?
> A: Yes, you should commit it. The presence of a `Gemfile.lock` in a gem’s repository ensures that a fresh checkout of the repository uses the exact same set of dependencies every time. We believe this makes repositories more friendly towards new and existing contributors. Ideally, anyone should be able to clone the repo, run `bundle install`, and have passing tests. If you don’t check in your `Gemfile.lock`, new contributors can get different versions of your dependencies, and run into failing tests that they don’t know how to fix.
> 
> Q: But I have read that gems should not check in the Gemfile.lock!
> A: The main advantage of not checking in your Gemfile.lock is that new checkouts (including CI) will immediately have failing tests if one of your dependencies changes in a breaking way. Instead of forcing every fresh checkout (and possible new contributor) to encounter broken builds, the Bundler team recommends either using a tool like [Dependabot](https://dependabot.com/) to automatically create a PR and run the test suite any time your dependencies release new versions.

TL;DR: **yes**, we should probably keep `Gemfile.lock` in Git, as it's what Bundler says to do and I'm inclined to agree with their reasoning. _However_ I really don't like having arch-specific deps listed in a file meant for everyone, so I'm going to see what I can do to fix that now.

SamBarker: I wouldn't sink to much time into it if thats the official advice. 

It also make me question my standard habit of `bundle update && bundle exec jekyll serve`

Maybe I should just be doing `bundle exec jekyll serve` or there is some other `bundle` command I should use rather than update...

gracegrimwood:
> Local builds also show a couple of warnings
> 
> ```shell
> asciidoctor: WARNING: available-filters/record-encryption/record-encryption.adoc: line 158: section title out of sequence: expected level 4, got level 5
> asciidoctor: WARNING: available-filters/record-encryption/record-encryption.adoc: line 158: section title out of sequence: expected level 4, got level 5
> ```
> 
> I presume those are also issues in the runtime source tree?

Yeah, those warnings are coming out of Asciidoctor (not the `jekyll-asciidoc` plugin, but the actual Asciidoctor text processor that the plugin invokes). When you build these docs with maven in the Kroxylicious repository you get the exact same warning:
```
[INFO] ----------------< io.kroxylicious:kroxylicious-parent >-----------------
[INFO] Building Parent POM 0.5.1                                         [1/24]
[INFO]   from pom.xml
[INFO] --------------------------------[ pom ]---------------------------------
[INFO] 
[INFO] --- asciidoctor:3.0.0:process-asciidoc (convert-to-html) @ kroxylicious-parent ---
Downloading from central: https://repo.maven.apache.org/maven2/org/asciidoctor/asciidoctor-maven-commons/3.0.0/asciidoctor-maven-commons-3.0.0.pom
Downloaded from central: https://repo.maven.apache.org/maven2/org/asciidoctor/asciidoctor-maven-commons/3.0.0/asciidoctor-maven-commons-3.0.0.pom (1.9 kB at 9.1 kB/s)
Downloading from central: https://repo.maven.apache.org/maven2/org/asciidoctor/asciidoctor-maven-commons/3.0.0/asciidoctor-maven-commons-3.0.0.jar
Downloaded from central: https://repo.maven.apache.org/maven2/org/asciidoctor/asciidoctor-maven-commons/3.0.0/asciidoctor-maven-commons-3.0.0.jar (19 kB at 86 kB/s)
[INFO] asciidoctor: DEBUG: overview.adoc: line 31: unknown style for literal block: a2s
[INFO] asciidoctor: DEBUG: overview.adoc: line 96: unknown style for literal block: a2s
[INFO] asciidoctor: DEBUG: overview.adoc: line 118: unknown style for literal block: a2s
[INFO] asciidoctor: DEBUG: overview.adoc: line 171: unknown style for literal block: a2s
[INFO] asciidoctor: WARN: available-filters/record-encryption/record-encryption.adoc: line 158: section title out of sequence: expected level 4, got level 5
[INFO] asciidoctor: DEBUG: custom-filters.adoc: line 288: unknown style for literal block: a2s
[INFO] Converted /Users/ggrimwoo/workspace/kroxylicious/docs/index.adoc
```
I think it's something to do with the `include::hashicorp-vault/service.adoc[leveloffset=3]` statement right above a level 4 heading ([lines 156-158](https://github.com/kroxylicious/kroxylicious/blob/main/docs/available-filters/record-encryption/record-encryption.adoc?plain=1#L156-L158)). Note that [further down](https://github.com/kroxylicious/kroxylicious/blob/main/docs/available-filters/record-encryption/record-encryption.adoc?plain=1#L195-L197) that file where we're including another file, we're including it at `leveloffset=3` above a level 3 heading.

Either way, it's not a fatal error, just a warning about some weird formatting.

gracegrimwood:
> Lets update the adoc to use rouge. One formatter to rule them all....

It would be worth checking if that's a supported option for the development docs, as we use legacy GitHub Pages (based on _very old_ versions of Jekyll and Ruby) for those, and installing custom gems is not an option with that. I know that the old version of Jekyll used by legacy GH Pages uses pygments as a code highlighter, and Asciidoctor only supports rouge versions >=2.x, but I don't know if there's any actual docs on specifically what code highlighters are supported.

If we can't use rouge with the development docs, we will have to modify this property in these files every time they get pushed into the website repo.

> Also worth doing, but I was expecting the website to maintain docs for each version not just a single copy of the current release docs. So /docs/0.5.0 & /docs/0.3.0 would work was my expectation. That said I'm fine with this PR just fixing the rendering and sorting versioning in a separate PR.

...I might be able to make that work. The difficulty will be that Jekyll isn't actually handling the AsciiDoc files, the `jekyll-asciidoc` plugin is offloading that rendering to Asciidoctor and largely just using Jekyll as a source of configs and styles before ultimately shoving Asciidoctor's rendered HTML into Jekyll's serving directory. That means Jekyll's whole convention around directories and how they translate to URI fragments only kind of applies to AsciiDoc files. Asciidoctor has its own logic around directories which overrides Jekyll's behaviour, which may mean adding extra metadata to the `.adoc` files to make this work.

Anyway, I'll see how difficult it is to do and either update this PR or create a new one for that.

Gemfile.lock
gracegrimwood: `bundle install` or just `bundle` is what I would use. Unless you're specifically playing with deps, I wouldn't use `bundle update`.

I'm usually running `bundle && bundle exec jekyll serve --livereload` (unless I'm changing things in the `Gemfile` or `_config.yml`, in which case I leave out `--livereload` and rebuild manually after each change because otherwise they don't always get picked up)

SamBarker:
Re asciidoc warnings, yeah thats fine. My only concern was if it was a change of behaviour between repos.

gracegrimwood:
I've created a PR against this branch to add the ability to render multiple docs versions and navigate to them sensibly, see here: https://github.com/gracegrimwood/kroxylicious.github.io/pull/1

k-wall:
nice work @gracegrimwood.    I notice the anchoring icon has been lost.  I value this feature.  It is really useful to be able to easily point someone at a specific part of the documentation during a support interaction.  Is it possible to restore it?

gracegrimwood:
> I notice the anchoring icon has been lost. I value this feature. It is really useful to be able to easily point someone at a specific part of the documentation during a support interaction. Is it possible to restore it?

I've added this to the multiple docs versions PR, and it should be live on the demo site now.

SamBarker (approved):
Awesome work @gracegrimwood !

LGTM in conjunction with https://github.com/gracegrimwood/kroxylicious.github.io/pull/1

feel free to squash them together or merge in turn.

PR #53, opened 2024-05-21
Add blog post for KSB 2024

SamBarker:


gracegrimwood:
From our discussion on Slack:

> Ok so the files under `./assets/images/` are RH logos for the site theme. I might move these to `./images/` because that's where the rest of the [static files](https://jekyllrb.com/docs/static-files/) for the site theme live.
> 
> As for the static files under `./assets/`: these appear to be largely for the use cases page, but there are some assets for blog posts in there which confuses things a bit. I'm fine with having static files for page content kept seperate from theme content (it means we don't have to dig through theme stuff to update pages and vice versa), but I think we should probably keep blog post files seperate because there's a good chance we're eventually going to end up with a lot of blog post files.
> 
> After looking at [what the Jekyll docs recommend](https://jekyllrb.com/docs/posts/#including-images-and-resources), my preference would be to either have a seperate `./blog_assets/` directory for blog post files, or to divide the existing ./assets/ directory into `./assets/blog/` and `./assets/pages/`.
> 
> However, if you're looking for a quick solution that doesn't involve restructuring the site and updating all the image links, you should put them in `./assets/`.

We should probably also document this in the README or similar to avoid any further confusion.

SamBarker (commented):
Everything appears to render properly locally and there are no 404 errors reported when browsing.

gracegrimwood (approved):
Tested and working locally, LGTM :)

PR #58, opened 2024-06-09
Fix typo/grammar

robobario:
Fix typo


SamBarker (approved):
Thanks for catching @robobario

PR #60, opened 2024-07-08
fix borken doc link

showuon:
The "documentation" link is directing to an 404 page: https://kroxylicious.io/overview/kroxylicious . Fix it. 
Question: Do we have a label with the latest version number? Otherwise, we need to remember to update it when the new version released.

### DIFF
--- a/overview.markdown
+++ b/overview.markdown
@@ -5,7 +5,7 @@ permalink: /overview/
 toc: true
 ---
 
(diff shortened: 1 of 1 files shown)


k-wall (approved):

k-wall:
thank you @showuon

k-wall:
> Question: Do we have a label with the latest version number? Otherwise, we need to remember to update it when the new version released. 

currently no, we need automation to reduce the toil associated with a new release.

PR #63, opened 2024-08-09
Schema enforcement use-case docs

k-wall:
preparing use-case docs ready for 0.7.0

Please don't merge this PR until 0.7.0 is released.


_use_cases/schema_validation_and_enforecement.markdown
```diff
@@ -0,0 +1,84 @@
...
+* _SyntacticallyCorrectJson_ validation ensures the producer is producing messages that contain syntactically valid
+  JSON.
+
+Validations may be configured on a per-topic basis.
+
+| ![image]({{'/assets/pages/images/schema_validation_solution.png' | absolute_url}}){:width="100%"} |
+|:-----------------------------------------------------------------:|
+|    *Proxy configured for schema enforcement*     |
+
+
```
k-wall, line 83: I'd like to add a sequence diagram here to explain the error flow.

_use_cases/schema_validation_and_enforecement.markdown
```diff
@@ -0,0 +1,84 @@
...
+The filter intercepts the produce requests sent from producing applications and subjects them to validation. If
+validation fail, the product request is rejected.  The producing application receives an error response.  The broker
+does not receive the rejected records.  In this way, one can organize that a poison message never enters the system,
+even if producing applications are misconfigured.
+
+The filter currently supports two types of validation:
+
+* _Schema validation_[^1] validates the content of the record against a schema.  
+* _SyntacticallyCorrectJson_ validation ensures the producer is producing messages that contain syntactically valid
```
k-wall, line 74: Maybe we shouldn't mention `SyntacticallyCorrectJson` in the use-case documentation at all.   I think it was really a POC that helped us prove the idea.

_use_cases/schema_validation_and_enforecement.markdown
```diff
@@ -0,0 +1,84 @@
...
+|:-----------------------------------------------------------------:|
+|    *Problem: Poison message leading to consumer crash*     |
+
+
+##### Schemas to the rescue
+
+To overcome the problem, the Kafka **client** ecosystem supports _schemas_.  Schemas provide a programmatic description
+of the message format.
+
+Producers use a schema to ensure that they produce messages that adhere to the agreed message format.  The consumer
```
SamBarker, line 29: My mental model was always
> producers use a schema to advertise the format of the messages they (are supposed to) produce

Which is where a proxy would step in to ensure the compliance with the schema.

_use_cases/schema_validation_and_enforecement.markdown
SamBarker: Its one of those things where it helps the unstructured JSON crowd. So if we are taking a binary view and pushing schemas everywhere then I don't think there is any value in mentioning it here.  I'd be quite happy with something like 
> The filter currently supports two modes of operation
> 1. _Schema validation_[^1] validates the content of the record against a schema.   - For topics which have an entry in the centralised schema registry
> 2. _SyntacticallyCorrectJson_ validation ensures the producer is producing messages that contain syntactically valid JSON - For topics which do not have registered schemas.  

Having mentioned the 2nd option I don't think we need to go much further.

assets/pages/images/schema_validation_problem.png
SamBarker, line 1: I like the diagram but the contrast between the happy and emoji and the hand drawn sad fave is a bit jarring. Could we replace the sad face with something else?

SamBarker (approved):
Nice work @k-wall

k-wall:
@robobario would you mind reviewing too?

_layouts/use-cases.html
```diff
@@ -9,6 +9,9 @@
             <div class="card-title display-6 mt-3 mb-4">Use Cases</div>
             <div class="row g-0">
                 <div class="col-auto">
+                  {% for use_case in site.use_cases %}
```
k-wall, line 12: There's probably a better way of achieving this :)

k-wall:
thank you for the feedback @ddaviesbrackett-dwp

_use_cases/schema_validation_and_enforecement.markdown
```diff
@@ -0,0 +1,95 @@
...
+The Kroxylicious Record Validation Filter provides a solution to the problem. 
+
+The filter intercepts the produce requests sent from producing applications and subjects them to validation. If
+validation fail, the product request is rejected and the producing application receives an error response.  The broker
+does not receive the rejected records.  In this way, one can organize that a poison message never enters the system,
+even if producing applications are misconfigured.
+
+The filter currently supports two modes of operation:
+
+1. Schema validation[^1] validates the content of the record against a schema. Use this for topics which have an entry in
```
robobario, line 71: We are also enabling another potential usecase, the user can use a schema from a Registry to validate json blobs being pushed to kafka by a client with no knowledge of Schemas. So we can use technology that is not apicurio integrated, but still prevent invalid JSON from entering the topic.

_use_cases/schema_validation_and_enforecement.markdown
```diff
@@ -0,0 +1,95 @@
...
+Producer applications call the registry to retrieve the schema for the messages they need to send.
+
+Consuming applications use information embedded in the incoming message to be able to retrieve the right schema from
+the registry.  This allows the consumer to decode the message.
+
+| ![image]({{'/assets/pages/images/schema_validation_schema_registry.png' | absolute_url}}){:width="100%"} |
+|:-----------------------------------------------------------------:|
+|    *Kafka applications making use of a Schema Registry*     |
+
+#### So, what's the problem?
```
robobario, line 49: Maybe we should also point out that maybe not all clients are schema-aware, we may have technology that wants to emit JSON without awareness of how to encode apicurio identifiers into the message. The schema validation check can enforce the schema and prevent invalid messages from these clients entering the topic.

robobario: might just muddy the usecase though

k-wall: I think we should go with it as it stands.  I'm sure we'll want to improve the use-case documentation as we grow the feature.

_layouts/use-cases.html
k-wall: @gracegrimwood do you know the right way to achieve this?  The other option would be to turn the use-case button into a drop down menu.

_use_cases/schema_validation_and_enforecement.markdown
```diff
@@ -0,0 +1,95 @@
+---
+name: Schema Validation and Enforcement
+---
+
+### Schema Validation and Enforcement
+
+#### Why
+
+In Kafka, producing applications transfer messages to consuming applications.  In order for messages to be transferred successfully
```
katheris, line 9:
```suggestion
In Apache Kafka, producing applications transfer messages to consuming applications.  In order for messages to be transferred successfully
```

_use_cases/schema_validation_and_enforecement.markdown
```diff
@@ -0,0 +1,95 @@
...
+producer and consumer need to agree on the format of the message being transferred. If there is a mismatch between the format sent
```
katheris, line 10:
```suggestion
producers and consumers need to agree on the format of the message being transferred. If there is a mismatch between the format sent
```

_use_cases/schema_validation_and_enforecement.markdown
```diff
@@ -0,0 +1,95 @@
...
+by the producer and the expectations of the consumer, problems will result.  The consuming application could take the wrong business
+action or the consuming application may fail leading to complete failure of the system.  This problem is known as a *poison message*
```
katheris, line 12:
```suggestion
action or the consuming application may fail to process the message at all, leading to complete failure of the system.  This problem is known as a *poison message*
```

katheris (approved):
A couple of small suggestions, but otherwise this looks good to me.

_layouts/use-cases.html
gracegrimwood: What's the goal? To be able to navigate _to_ each use case, or to be able to navigate _within_ each use case? If you look at what's been done for the [docs layout](https://github.com/kroxylicious/kroxylicious.github.io/blob/main/_layouts/docs.html), with the internal navigation in a sidebar, does that sort of achieve what you're after?
If we want to repurpose that layout for the use cases we'll have to duplicate it, the docs layout is set up to process AsciiDoc rather than markdown.

Otherwise if you just want buttons that take you to each use case, there's always the [Slugify filter](https://jekyllrb.com/docs/liquid/filters/). Something like:
```
{% for use_case in site.use_cases %}
    {% include my_cool_button.html label=use_case.name %}
{% endfor %}
```
and then in `_includes/my_cool_button.html` you'd have something like this ([bootstrap button docs](https://getbootstrap.com/docs/5.3/components/buttons/)):
```
<button type="button" class="btn btn-primary" href="#{{ name | slugify }}">{{ name }}</button>
```
and then you just arrange those buttons however you'd like by adding structure around them with [grid components](https://getbootstrap.com/docs/5.3/layout/grid/) in `_layouts/use_cases.html`. The name property already exists in the existing use case front matter so it's just a matter of passing it through slugify to get something URL-safe.

k-wall: > What's the goal? To be able to navigate to each use case, or to be able to navigate within each use case? 

really the former.  currently, if you click on the use-case navigation you see "record encryption".   There's nothing obvious that  lets the reader know there's further use-cases below.  I vomited a TOC onto the page (sorry) as a quick and dirty solution.

Unfortunately, I don't have the jekyll skills (or the available bandwidth right now to learn).. so if you'd be able to make a suggestion, that would be appreciated.

PR #74, opened 2024-09-12
bump bootstrap to v5.3.3

gracegrimwood:
Does what it says on the tin, bumps Bootstrap to the latest [version (v5.3.3)](https://github.com/twbs/bootstrap/releases/tag/v5.3.3).

### DIFF
--- a/bootstrap_setup.sh
+++ b/bootstrap_setup.sh
@@ -2,6 +2,10 @@
 
 export BOOTSTRAP_VERSION="5.3.3"
 
+#Remove any existing Bootstrap sources
+rm -f "v$BOOTSTRAP_VERSION.zip"
+rm -rf ./_sass/bootstrap
+
 # download the full Bootstrap 5 sources (including JS)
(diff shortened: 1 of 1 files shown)


SamBarker:
Decided to be annoying and actually test this. Admittedly on my existing checkout which is already bootstrapped, it downloads 5.3.3 as expected but
```
mkdir: cannot create directory ‘./_sass/bootstrap’: File exists
mv: cannot overwrite './_sass/bootstrap/scss/scss': Directory not empty
```
Is that to be expected? Should bootstrap be idempotent?

gracegrimwood:
@SamBarker The `bootstrap_setup.sh` script assumes that the directory does not already exist. For local development this is not necessarily an accurate assumption, but for the GitHub Actions build it works fine... It would be trivial to add a check or just have it remove the existing bootstrap files itself, I can add that today.

SamBarker (approved):
LGTM

Tested locally and things work :D

PR #75, opened 2024-09-13
Create 2024-09-13-junit-ext-and-kafka4.md

k-wall:
Blog post about https://github.com/kroxylicious/kroxylicious-junit5-extension/issues/389

### DIFF
--- a/_posts/2024-09-13-junit-ext-and-kafka4.md
+++ b/_posts/2024-09-13-junit-ext-and-kafka4.md
@@ -1,0 +1,16 @@
+---
+layout: post
+title:  "Moving Kroxylicious Junit5 Extension to Kafka 4.0"
+date:   2024-09-13 00:00:00 +0000
+author: "Keith Wall"
+author_url: "https://www.github.com/k-wall"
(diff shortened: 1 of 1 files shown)


SamBarker (approved):
Think the post should make it clearer that were are only really discussing in memory usage but will continue to support zookeeper and older broker clusters in vm mode

k-wall:
> Think the post should make it clearer that were are only really discussing in memory usage but will continue to support zookeeper and older broker clusters in vm mode

If it turns out no one cares are Zookeeper at all, why not kill both container and in-VM paths?  Fewer lines of code to maintain.

SamBarker:
> > Think the post should make it clearer that were are only really discussing in memory usage but will continue to support zookeeper and older broker clusters in vm mode
> 
> 
> 
> If it turns out no one cares are Zookeeper at all, why not kill both container and in-VM paths?  Fewer lines of code to maintain.

No particular reason but that was my take away from the call, we were only going to remove in memory support not container.

k-wall:
Let's see what feedback comes back.

PR #81, opened 2024-11-06
Add missing "old docs" warning messages to previous release docs pages

gracegrimwood:
The docs pages for versions up to and including v0.5.0 have a message at the top which reads `This documentation is for an older release of Kroxylicious, and may not reflect current functionality.` This message is missing from the docs for v0.5.1, v0.6.0, and v0.7.0 (v0.8.0 is the current release so shouldn't have this message), so this PR is just to add that missing message to those pages.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -109,6 +109,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-22
+  arm64-darwin-23
   java
   x64-mingw32
   x86_64-darwin-22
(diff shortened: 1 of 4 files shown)


SamBarker (approved):
I don't think I'd really appreciated that adding the warning was a manual process.

LGTM

robobario (approved):

k-wall:
> I don't think I'd really appreciated that adding the warning was a manual process.

As a minimum, we should document that we need to add 'old docs' to the previous release when releasing a new one.  Ideally, it should be automated.

PR #100, opened 2025-02-05
Accessibility Fixes

### LINKED ISSUE #99: Ensure Kroxylicious website is accessible
Parts of the kroxylicious.io website are not compatible with screen readers or compliant with accessibility standards. We should fix this, and perhaps also add some script to GitHub Actions to prevent changes that break accessibility in future.

gracegrimwood:
Using the [HTML Codesniffer bookmarklet](https://github.com/squizlabs/HTML_CodeSniffer), I'm going through the website and trying to fix the accessibility issues mentioned in #99. Some of these will likely be in the docs which might be a more complicated fix (we'll have to update all the versions of the docs in the website repo, and then also make the same changes in the main Kroxylicious repo).

Closes #99

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -29,14 +29,14 @@
                         </div>
                         <div class="w-100 my-2">
                             <a class="icon-link me-1" href="https://github.com/kroxylicious/kroxylicious">
-                                <img class="img-fluid krx-icon" src="{{ '/assets/theme/images/github-mark.svg' | absolute_url }}"/>
+                                <img class="img-fluid krx-icon" src="{{ '/assets/theme/images/github-mark.svg' | absolute_url }}" alt="GitHub logo"/>
                             </a>
                         </div>
                     </li>
                 </ul>
             </div>
             <div class="col-6 col-lg-2 offset-lg-2 mb-3 text-lg-end">
-                <h5>Links</h5>
+                <h3 class="fs-5">Links</h3>
                 <ul class="list-unstyled">
                     <li class="mb-2">
                         <a class="text-body link-underline-dark link-underline-opacity-0 link-underline-opacity-10-hover" href="{{ '/' | absolute_url }}">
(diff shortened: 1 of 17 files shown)


gracegrimwood:
I think I've got everything that matters, but I would really appreciate it if someone else could double check.

gracegrimwood:
The only thing I couldn't fix is the duplicate IDs of the table of contents on the use cases page, because [jekyll-toc](https://github.com/toshimaru/jekyll-toc) is generating a TOC for each use case individually, and doesn't let us configure any kind of dynamic ID for them.

SamBarker (approved):
LGTM 

I'm no longer seeing any violations reported by the IBM checker.

gracegrimwood:
The tables in the AsciiDoc files need to either have headers or be labelled with `role="presentation"`, but aside from that everything should be fixed.

SamBarker:
>The tables in the AsciiDoc files need to either have headers or be labelled with role="presentation", but aside from that everything should be fixed.

How do we add `role="presentation"` to the AsciiDoc files? I appreciate its a sledge hammer but my instinct is thats it is the simpler option

gracegrimwood:
> How do we add role="presentation" to the AsciiDoc files?

That's a really good question that I don't have the answer to. I suspect it's something we put in the AsciiDoctor config, but I would check that with someone who knows AsciiDoc better than me 😄

SamBarker:
> > How do we add role="presentation" to the AsciiDoc files?
> 
> That's a really good question that I don't have the answer to. I suspect it's something we put in the AsciiDoctor config, but I would check that with someone who knows AsciiDoc better than me 😄

Hmm thats awkward as you are my goto expert. I'm going to merge this as is and we can revisit  the ascii doc later.

SamBarker (approved):
All violations outside the docs section are fixed.

PR #110, opened 2025-03-26
DRAFT: Restructure docs navigation

### LINKED ISSUE #106: Restructure documentation navigation
We're working on an Kubernetes operator and consequently are likely to end up with two (or three) separate guides, rather than the current monolithic guide. These guides will be tied to a specific Kroxylicious release. This will mean we need to restructure the website's navigation to support multiple guides. 

One way to do this would be to have a per-version "docs landing page", which then has links to the different sets of documentation applicable to that release (and maybe the Javadoc too -- see #45). The "Docs" item in the header would remain the same (expanding into a list of versions) but each version would link to the corresponding landing page. We'd need to decide how to handle the existing URLs (e.g. the current monolithic docs live at https://kroxylicious.io/docs/v0.10.0/, which is the intuitive URL for the per-release docs landing page).

gracegrimwood:
Fixes #106

Work in progress to add new docs subsections dynamically to dropdown nav (done), ~~and move old docs versions to an "older versions" index page and out of the main nav menu (in progress)~~ <- moving this to a second PR.

Trying to simplify things and avoid clutter.

Also includes some updating of "this is an older version" warnings on old docs versions I hadn't got around to adding it to. Probably we should find a better (i.e. not manual) way to do that in future but it's of very low importance.

### DIFF
--- a/_data/kroxylicious.yml
+++ b/_data/kroxylicious.yml
@@ -3,6 +3,9 @@ versions:
     url: '/kroxylicious'
   - title: 'v0.11.0'
     url: '/docs/v0.11.0/'
+    subsections:
+      - title: 'Proxy Guide'
+        url: '/docs/v0.11.0/kroxylicious-proxy/'
   - title: 'v0.10.0'
     url: '/docs/v0.10.0/'
   - title: 'v0.9.0'
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -47,11 +47,29 @@
             <a class="nav-link dropdown-toggle krx-nav-link py-2 px-0 px-lg-4" data-bs-toggle="dropdown" href="#" role="button" aria-expanded="false">Docs</a>
             <ul class="dropdown-menu">
               {% for version in site.data.kroxylicious.versions %}
+              {% if version.subsections %}
+              <li><hr class="dropdown-divider"></li>
+              <li><h6 class="dropdown-header">{{ version.title }}</h6></li>
+              {% endif %}
               <li>
                 <a class="dropdown-item {% if page.url == version.url %}active{% endif %} krx-nav-link py-2 px-0 px-lg-4" href="{{ version.url }}">
+                  {% if version.subsections %}
+                  Index
+                  {% else %}
                   {{ version.title }}
+                  {% endif %}
                 </a>
               </li>
+              {% if version.subsections %}
+              {% for subsection in version.subsections %}
+              <li>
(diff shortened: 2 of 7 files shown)


gracegrimwood:
### Some screenshots of the changes so far:

#### New subsections in dropdown nav
Screenshot 2025-03-26 at 18 21 29

To be honest I'm still not too happy with the look of this, but I tried several other options (indenting the subsection text using icons like arrows, dots, and chevrons to indicate subsections, using no headings on versions with subsections, using headings on versions regardless of subsections, etc etc) but this is the best looking of all of them I think. Part of the issue is that the menu is quite long and very narrow already which doesn't look great, and no matter what we do the menu will still be longer when subsections are added to it. I'm hoping that moving old versions of the docs out of this menu will help, so watch this space.

#### Ensure short pages (like the v0.11.0 docs index) are fit to the viewport height so there's no whitespace under the footer
Screenshot 2025-03-26 at 18 21 55

This was something I spotted after the last release, that the footer ended up floating halfway up the page which looked a bit crap. It happens on other pages too, like short blog posts and stuff, so I modified the base layout for all pages to ensure that they'll always be at least the height of the viewport (the size of the screen that's viewing the page, minus any browser or system UI space) and the footer will always be at the bottom of the screen.

SamBarker:
Personally I'd separate the two things into different PR's. I think this can be merged right now! 

Ideally we'd collapse the list of older versions under a `older releases` node in the menu as well, but I don't think thats essential.

gracegrimwood:
> Ideally we'd collapse the list of older versions under a `older releases` node in the menu as well, but I don't think thats essential.

Yeah we could specify a limit of versions in the menu and collapse everything else, that's doable. ([Link is here to the Liquid syntax docs](https://shopify.github.io/liquid/tags/iteration/#:~:text=for%20(parameters)-,limit,-Limits%20the%20loop) for my own reference on doing this.) Would have to figure out how bootstrap would cope with a collapse nested inside a dropdown menu nested inside a navlist but I can't see it being too much of a problem.

Will take this PR out of draft state now and push up a second PR with that change once I've made it.

gracegrimwood:
Actually, I've had another thought about collapsing older releases within the menu — when enough older releases are in that menu, it'll run off the page. That's not a problem on mobile, where this list is scrollable, but I believe (going to go double-check this now) that on desktop it isn't scrollable and that'll mean older versions become inaccessible via nav for most people.

SamBarker (approved):
> Actually, I've had another thought about collapsing older releases within the menu — when enough older releases are in that menu, it'll run off the page

Thats even more of an issue for the current model where they are all at the top level but thats a problem for the future.

Lets merge what we have and iterate on the rest separately.

PR #111, opened 2025-03-27
Fix formatting of docs

gracegrimwood:
## Summary:

- Adds a script in `_includes/asciidoctor-scripts/icons.html` to change some HTML classes on the release docs pages
- Adds some CSS to `_sass/kroxylicious.scss` to fix some missing styles on the release docs pages
- Moves the script to fix sectanchors out of the docs layout file and into `_includes/asciidoctor-scripts/sectanchors.html` to keep things neat

### Problem:

The release docs are currently not formatted correctly on the website. Compare the development docs:

Screenshot 2025-03-27 at 13 27 21

With the latest release docs:

Screenshot 2025-03-27 at 13 27 04

This is because the default Asciidoctor stylesheet is missing from the docs pages, as it's trying to use the ones Jekyll provides for the rest of the site. However the styles used by the rest of the site were not written with the Asciidoctor stylesheet in mind, so they do not apply completely and some things are missing.

### Other Issues:

To complicate matters further, [Asciidoctor](https://docs.asciidoctor.org/asciidoc/latest/macros/icons-font/) uses [FontAwesome](https://fontawesome.com/) (an SIL OFL licensed project with a free and paid tier) while the website uses [Bootstrap Icons](https://icons.getbootstrap.com/) (MIT licensed OSS, completely free), and this default icon library used by Asciidoctor cannot be changed (you can use text instead of icons but it looks weird so I elected not to do that). These two icon libraries also aren't 1:1 in terms of what icons they have (or allow you to use, in the case of FontAwesome's free tier) and they also use completely different element class names to apply icon images and styles.

### Fix:

To fix this without adding yet another font library (and associated licenses and such) to the site, I've added a script to the docs layout (via an `_includes`) which transforms these classes from the FontAwesome ones applied by Asciidoctor to my best approximation of an equivalent class in Bootstrap Icons. The script also applies some other Bootstrap Sass classes (to turn these [Asciidoc admonitions](https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/) into [Bootstrap alerts](https://getbootstrap.com/docs/5.3/components/alerts/)) and a few classes specific to the styling of our site (all the stuff prefixed with `krx-`).

I've then recreated and tweaked the relevant missing parts of the default Asciidoctor stylesheet we're using on the development docs, but inside the website's `_sass/kroxylicious.scss` main stylesheet so they should be accessible all over the site (just in case any Asciidoc content ever makes its way outside `/docs`).

### DIFF
--- a/_sass/kroxylicious.scss
+++ b/_sass/kroxylicious.scss
@@ -42,6 +42,12 @@ a {
   color: $kroxy-dark-green;
 }
 
+// Override default Bootstrap icon sizing and positioning for alerts
+.alert .bi {
+  font-size: xx-large!important;
+  vertical-align: middle!important;
+}
+
 .nav-item > .dropdown-menu {
   min-width: 200px;
 }
@@ -161,7 +167,7 @@ div.highlighter-rouge {
   }
 }
 
-// Anchor Links
+// [DOCS] Anchor Links
 
 h2>a.anchor>span, h3>a.anchor>span, h4>a.anchor>span, h5>a.anchor>span, h6>a.anchor>span {
   font-family: bootstrap-icons!important;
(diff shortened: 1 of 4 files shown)


gracegrimwood:
For reference, here's what that part of the docs shown above looks like with these changes:
Screenshot 2025-03-27 at 19 55 34

Much better, I think.

_sass/kroxylicious.scss
```diff
@@ -42,6 +42,12 @@ a {
   color: $kroxy-dark-green;
 }
 
+// Override default Bootstrap icon sizing and positioning for alerts
+.alert .bi {
+  font-size: xx-large!important;
+  vertical-align: middle!important;
+}
```
gracegrimwood, line 49: This _does_ change the size and layout of icons within alerts on all other pages, but I checked and it all looks fine (better, in my opinion 😉)

SamBarker (approved):
Based on the screenshots 

LGTM

SamBarker:
Relatedly @gracegrimwood do you know of any way to make a rendered/served copy of a PR copy available?

gracegrimwood:
> Relatedly @gracegrimwood do you know of any way to make a rendered/served copy of a PR copy available?

@SamBarker the way I've done it in the past is to create a branch for demo-ing the PR (specifically [this one](https://github.com/gracegrimwood/kroxylicious.github.io/tree/demo-on-gh-pages)) which has a specific GitHub workflow for rendering and serving that branch at my own GitHub subdomain (gracegrimwood.github.io, which I don't otherwise use).

We could probably set up a workflow to do this at some Kroxylicious subdomain for each PR, but you'd run into issues of multiple PRs overwriting each other's demos if you used a fixed subdomain, or the potential of having a bunch of old demo bits left hanging around if you used some dynamic subdomain (like `$PR_NUMBER`.kroxylicious.io or something). We wouldn't be able to have everyone use their own GH subdomain for this (instead of the Kroxylicious one) because people may have sites already at those addresses and we *really* don't want to be overwriting other people's stuff arbitrarily.

The easiest and simplest option is either PR authors post screenshots of their changes (like I've done above) or reviewers pull down the code and render it locally, thus avoiding all of the complications of trying to do those things up on the internet.

SamBarker:
> > Relatedly @gracegrimwood do you know of any way to make a rendered/served copy of a PR copy available?
> 
> @SamBarker the way I've done it in the past is to create a branch for demo-ing the PR (specifically [this one](https://github.com/gracegrimwood/kroxylicious.github.io/tree/demo-on-gh-pages)) which has a specific GitHub workflow for rendering and serving that branch at my own GitHub subdomain (gracegrimwood.github.io, which I don't otherwise use).
> 
> We could probably set up a workflow to do this at some Kroxylicious subdomain for each PR, but you'd run into issues of multiple PRs overwriting each other's demos if you used a fixed subdomain, or the potential of having a bunch of old demo bits left hanging around if you used some dynamic subdomain (like `$PR_NUMBER`.kroxylicious.io or something). We wouldn't be able to have everyone use their own GH subdomain for this (instead of the Kroxylicious one) because people may have sites already at those addresses and we _really_ don't want to be overwriting other people's stuff arbitrarily.
> 
> The easiest and simplest option is either PR authors post screenshots of their changes (like I've done above) or reviewers pull down the code and render it locally, thus avoiding all of the complications of trying to do those things up on the internet.

The reference to rendered was me wondering if we could build a bunch of static html and attach that to the pr but I don’t suppose that’s any easier to use than just pulling the PR and running locally

PR #114, opened 2025-04-07
Add thank you for 1password

robobario:


robobario:
not sure, I pulled it out of a press kit delivered as a download

PR #115, opened 2025-04-08
DRAFT: Docs Restructure Part 2

gracegrimwood:
Move old docs versions to an "older versions" archive page and out of the main nav dropdown menu.

New docs nav dropdown looks like this:
image

Archive page looks like this:
image

There's room in the archive page functionality to handle docs with subsections (like the v0.11.0 docs) and the addition of javadoc links if we elect to do that at some stage.

Currently in draft because I'll need to modify the release docs publishing pipeline to work with the changes to the site data structure before we merge.

### DIFF
--- a/docs/archive/index.markdown
+++ b/docs/archive/index.markdown
@@ -1,0 +1,29 @@
+---
+layout: docs
+title: Kroxylicious Documentation Archive
+permalink: /docs/archive/
+is_archive: true
+---
+{% assign earliest_documented_version = site.data.kroxylicious.older_versions | last %}
+{% capture earliest_docs_info %}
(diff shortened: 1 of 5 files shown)


docs/archive/index.markdown
```diff
@@ -0,0 +1,29 @@
+---
+layout: docs
+title: Kroxylicious Documentation Archive
+permalink: /docs/archive/
+is_archive: true
+---
+{% assign earliest_documented_version = site.data.kroxylicious.older_versions | last %}
```
SamBarker, line 7: Are we missing versions? Or is this just a belt and braces fix?

SamBarker (approved):
LGTM

Issue #124, opened 2025-04-28
Missing diagrams


tombentley:
The operator guide has a few diagrams which are not showing up on the website. For example [this section](https://kroxylicious.io/kroxylicious/kroxylicious-operator/#api_resources_used_by_the_kroxylicious_proxy) is supposed to have Figure 1 and Figure 2.

k-wall:
As I discovered in #140, it is the prefixing we have in the sources that is preventing images from being rendered properly.

```
-image::../_assets/monitoring-message-counters.svg["Conceptual diagram showing the downstream and upstream message metrics within the proxy, illustrating how they respond to message transit through it."]
+image::monitoring-message-counters.svg["Conceptual diagram showing the downstream and upstream message metrics within the proxy, illustrating how they respond to message transit through it."]
```

I'm wondering why we have the `./_assets/` part at all?  I know there is an asciidoc var `imagedir`.  Should that be specifying the location of the images?

robobario:
diagrams seem to have been sorted out somewhere along the line with various sweeping refactors

https://kroxylicious.io/documentation/0.14.0/html/kroxylicious-operator/

PR #140, opened 2025-06-23
Manually fix image locations with v0.13.0 docs

k-wall:
I'm not sure of the correct permanent fix, but I can see that the prefixing we do of image locations in the source asciidoc is messing up the publication on the website.

This is a temporary workaround.  We need to fix it properly.


franvila (approved):

k-wall:
@tombentley had raised this already https://github.com/kroxylicious/kroxylicious.github.io/issues/124

PR #141, opened 2025-06-27
Rework the website

tombentley:
* Change how `Documentation` works, add a `Download` to the navbar, and remove `Use cases` and `Quickstarts`.
* Remove the need for a drop-down menu enumerating historical versions by having a _documentation index page_ for each release. 
* The `Documentation` in the nav bar just links to the documentation for whatever is the current version (http://localhost:4000/documentation/). image
* Each release's documentation index page (e.g. http://localhost:4000/documentation/0.13.0/) links to an archive of all historical releases (http://localhost:4000/release-archives/). 
image
* The `Download` page enumerates assets in the release, and likewise links to the archive of all historical releases.
image
* `Use cases` is still accessible from the button on the hompage.
* `Quickstarts` are now accessible from the documentation.
* When accessing any download or documentation for a non-current version show a warning. e.g. 
image 
image
* The documentation and assets are driven declaratively from YAML files in `_data`.
* I have corresponding build changes for the [kroxylicious/kroxylicious](https://github.com/kroxylicious/kroxylicious) repo which renders the documentation and builds the documentation YAML for a release using maven and means that "releasing the docs" is a `cp -r ../kroxylicious/target/web/* .` + git push in a clone of the website repo.
* These changes make the website work better as the number of documentation guides and releases grows in the future.
* All the pre-0.13.0 documentation still lives in `/docs` and is rendered by the jekyll asciidoc plugin. Going forward the pre-rendered HTML docs will live under `/documentation`. Those URLs look like `/documentation/$VERSION/html/$GUIDE`. That should allow us to serve pdfs from `/documentation/$VERSION/pdf/$GUIDE.pdf` in the future.


_data/documentation/0_13_0.yaml
```diff
@@ -0,0 +1,38 @@
+docs:
```
tombentley, line 1: This is an example of the YAML which is used to drive the documentation index page. This is/will be generated from the `mvn` build in https://github.com/kroxylicious/kroxylicious, so the metadata about a piece of documentation lives beside that documentation.

_data/release.yaml
```diff
@@ -0,0 +1,117 @@
+# This object is used for the new-style /documentation and /download paths
+
+# The version number of the latest release
+latest: 0.13.0
+# All releases (in descending order)
+list:
```
tombentley, line 6: Probably we should split this up like I've done with the `_data/documentation`, because it's easier to add a self-contained YAML file when doing a release than to update an existing one.

documentation/0.13.0/html/developer-guide/index.html
```diff
@@ -0,0 +1,6 @@
+---
```
tombentley, line 1: These `index.html` files get generated by the maven build.

documentation/0.13.0/html/developer-guide/toc.html
```diff
@@ -0,0 +1,19 @@
+{% raw %}
```
tombentley, line 1: The content.html and toc.html also get generated by the maven build. So jekyll's asciidoc plugin is no longer used (except for the old docs which I've not attempted to convert in this PR).

documentation/0.14.0-SNAPSHOT/html/developer-guide/index.html
```diff
@@ -0,0 +1,6 @@
+---
```
tombentley, line 1: I know we're not actually going to publish SNAPSHOT documentation using this mechanism, but it was a useful exercise.

tombentley:
@k-wall I added support for `rank` in the per-document metadata, so we have explicit control over the order that documents are shown in on the index page. I also adopted your wording for the guide.

tombentley:
@k-wall I put the record encryption guide there because it's a guide. I hear what you're saying, and I agree we'll need something better when the number of docs grows, but for now I think this is good enough.  If we can merge this then we're in a much better position to iterate.

It not entirely clear to me what the best way of categorising the docs would be. We have:

* document type (quickstart, guide, and api reference)
* persona (ops vs dev)
* component type (proxy, operator, filter)

I think depending on what the user wants to achieve each of these is a reasonable way group or filter.

I had a few other ideas, which would also help:

* Icons (e.g. Kroxy for the proxy, the Kube logo for the operator, Duke for developer stuff)
* Do we want grouping or interactive filtering, based on the document tags. That would take a little bit of JS, but avoids imposing a categorisation on the user.

I've not implemented any of these yet because this is already a big PR.

robobario:
+1 to refining download groupings later.

One thing I'm wondering is if the latest quickstarts could be added to the frontpage somewhere more prominent. Another button in the big central bit like "Get Started!"?

I wonder if use-cases and what is it really should be split, the use cases justify the proxy model, maybe they should be a single page (but we would have to worry about existing google indexes that land on `use-cases`)

robobario:
you're also getting pinged by the automation for DCO signoff @tombentley

k-wall:
LGTM.  All steps in the right direction, thanks @tombentley

k-wall:
> I'd also like to version them and keep them in code repo, so we can't easily update them as we refactor.

My thoughts exactly.

k-wall (approved):

PR #144, opened 2025-07-01
Upgrade ruby to 3.4.4 and gems to latest

robobario:
While trying to Dockerize this I've hit on issues with the gems and lockfile, I thought it would be good to get everything freshened up and reset the Gemfile.lock from scratch (rather than accumulating stuff on top of the old lockfile). If someone could run it on OSX and contribute the changes to this branch that'd be helpful.

I also hardened up the Gemfile versions, specifying what version we want of each plugin.

Also have removed jruby/windows support pending anyone wanting them.

This has built successfully on my fork:

https://github.com/robobario/kroxylicious.github.io/actions/runs/16010238274
https://robobario.github.io/kroxylicious.github.io/


robobario:
We don't use github to build our pages. You can configure GitHub to run the jekyll build for you, you point it at a branch and github will build it for you, using jekyll and the versions specified on that page.

In our [workflow](https://github.com/kroxylicious/kroxylicious.github.io/blob/main/.github/workflows/jekyll-gh-pages.yml), we build from Jekyll sources to static html etc using a custom ubuntu step where we can do interesting things like install a2s. Then upload the resulting `_site` directory as an artifact and the action deploy's it to pages.

We already build with Jekyll 4.3.2, where github uses 3.10.0

tombentley:
Cool, that works (with the same SASS deprecation warnings which we already had)!

robobario:
think I'll merge now and an OSX user can contribute their lockfile changes separately

PR #145, opened 2025-07-02
GitHub build with docker image

robobario:
This means the production site will be built with the same Dockerfile that we can use locally for development. Bringing the environments closer together.


robobario:
something has gone wrong here

PR #146, opened 2025-07-02
Build the website using the same Dockerfile in the github action

robobario:
This brings the local development and github actions builds closer together. Using the same Dockerfile in both places to build the static content.

1. we build the image and cache the layers it to gha, this means the build step should usually be very fast.
2. we do not copy in all the sources into the image because we mount the sources in locally and in the workflow. this makes the image very cacheable.
3. moves a2s into /usr/local/bin so that the github user id can access it (previously it was under /root).

### DIFF
--- a/.github/workflows/jekyll-gh-pages.yml
+++ b/.github/workflows/jekyll-gh-pages.yml
@@ -25,29 +25,41 @@ jobs:
   # Build job
   build:
     runs-on: ubuntu-latest
+    services:
+      # we spin up a registry to share the image built by build-push-action with the later docker run step
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Setup Ruby
-        uses: ruby/setup-ruby@v1
-        with:
-          ruby-version: '3.4.4' # Not needed with a .ruby-version file
-          cache-version: 0 # Increment this number if you need to re-download cached gems
       - name: Setup Pages
         uses: actions/configure-pages@v3
-      - name: Install rake
-        run: gem install rake
-      - name: Update Bundler and Setup Jekyll
-        run: bundle install && bundle update --bundler
-      - name: Setup Go
-        uses: actions/setup-go@v5
-      - name: Setup ASCIIToSVG
-        run: go install github.com/asciitosvg/asciitosvg/cmd/a2s@latest
-      - name: Download Bootstrap 5 SCSS Sources
-        run: ./bootstrap_setup.sh
+      
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          # this is required for the subsequent build to be able to push to the registry on localhost:5000
+          driver-opts: network=host
+
+      - name: Build Docker image
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          push: true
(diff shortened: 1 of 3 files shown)


.github/workflows/jekyll-gh-pages.yml
```diff
@@ -35,19 +40,29 @@ jobs:
...
-        uses: actions/setup-go@v5
-      - name: Setup ASCIIToSVG
-        run: go install github.com/asciitosvg/asciitosvg/cmd/a2s@latest
-      - name: Download Bootstrap 5 SCSS Sources
-        run: ./bootstrap_setup.sh
+      
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver-opts: network=host
```
robobario, line 47: this is required to push to localhost:5000 and use the registry we spin up for the job

.github/workflows/jekyll-gh-pages.yml
```diff
@@ -35,19 +40,29 @@ jobs:
...
       - name: Build with Jekyll
         # Outputs to the './_site' directory by default
-        run: bundle exec jekyll build --baseurl "${{ steps.pages.outputs.base_path }}"
+        run: |
+          docker run \
+            --rm \
+            -u "$(id -u):$(id -g)" \
+            -v "$(pwd):/site" \
+            localhost:5000/kroxy-jekyll:latest \
+            bash -c 'eval "$(rbenv init -)" && cp -r /css/_sass/bootstrap /site/_sass/ && bundle exec jekyll build --baseurl "${{ steps.pages.outputs.base_path }}"'
```
robobario, line 65: the `--baseurl "${{ steps.pages.outputs.base_path }}` isn't actually doing anything because we don't have an `id: step` on the earlier configure pages step.

It's kind of moot since the pages are generally pretty broken if you deploy them on a fork, the docs links don't appear to honour base_url and some links like the quickstarts in the top nav go straight to kroxylicious.io which I think is because it's configured into _config.yml as the url. Needs a whole other PR to try and get it set up to work on a fork better.

robobario: (existing issues with running on forks)

.github/workflows/jekyll-gh-pages.yml
tombentley: This is probably worthy of a comment in the file.

.github/workflows/jekyll-gh-pages.yml
tombentley: What do you mean by "deploy on a fork"?

tombentley (approved):
This seems reasonable to me.

.github/workflows/jekyll-gh-pages.yml
robobario: I have my fork set up to deploy a branch to github pages

https://robobario.github.io/kroxylicious.github.io/

because of the way the job is configured with site.url configured to `kroylicious.io` and it not being aware of the extra `kroxylicious.github.io` it looks wrong.

with a couple of changes we could make it work neatly on a fork, it helps when testing changes to the github workflows

robobario: edit: sorry, I mean it deploys the same way as on this repo. I push to main on my fork and the action deploys to pages.

PR #149, opened 2025-07-02
Kill of unused data

tombentley:
`grep -Fr data.kroxylicious | grep -Fv '.latestRelease'` shows now hits, so we know the `versions` property is now unused.
Let's remove it to avoid future confusion.


robobario (approved):
LGTM, can't find any refs to `site.data.kroxylicious.versions` except in git history

PR #151, opened 2025-07-02
Sort releases

robobario:
This requires each release file to contain a rank property which orders by the release version. We need to leftpad the version components with zeroes to keep the ordering lexicographical.

It also adds a `version` property to the release YAML since I couldn't find a way to sort an array of arrays. And working backwards from `rank` would be painful.

This removes the monstrosity

### DIFF
--- a/_data/release/0_10_0.yaml
+++ b/_data/release/0_10_0.yaml
@@ -1,3 +1,5 @@
+version: 0.10.0
+rank: 0000.0010.0000
 assetBaseUrl: https://github.com/kroxylicious/kroxylicious/releases/tag/v$(VERSION)
 assets:
   - name: Proxy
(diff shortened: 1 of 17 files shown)


tombentley:
I don't love the duplication of version numbers in _data filenames and two properties therein. Not really sure which is the worse evil: that duplication or the monstrosity. Choose your cup of liquid poison, I suppose. But I guess the monstrosity is worse.

However, if we merge this I'd like to make corresponding changes to the `webify.java` script in my PR in the other repo. Right now all the YAML in `_data/documentation` is basically duplicating the frontmatter for docs which have been migrated to `/documentation`, and the `webify.java` is what keeps that in sync for all _new_ documentation. It would be nice if we could eventually get rid of `_data/documentation`, or at least slim it down, and rely on the frontmatter.

k-wall:
I could live with rank.

k-wall (approved):

robobario:
closing, merged the alternative using a custom liquid filter to do the sort

PR #154, opened 2025-07-03
Upgrade bootstrap to 5.3.7

robobario:
Fixes some deprecation warnings spewed by jekyll (but not all):

```
Deprecation Warning [mixed-decls]: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.
```

### DIFF
--- a/bootstrap_setup.sh
+++ b/bootstrap_setup.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-export BOOTSTRAP_VERSION="5.3.3"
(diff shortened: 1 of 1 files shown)


tombentley:
Did we have any instances of the kinds of declarations the deprecation warning was about, or was it getting displayed even in their absence?

robobario:
the declarations it is warning about are in the bootstrap scss we are building

PR #156, opened 2025-07-03
Use a replacement token for the per-version documentation root

robobario:
If a documentation path contains $(DOC_BASE_FOR_VERSION) we replace this with `/documentation/$(version)`

Why

Previously, if the path did not start with "http://" or "https://" or "/", we inferred that it is relative to the per-version documentation dir. A path of `html/123` for version 0.13.0 would be expanded to: documentation/0.13.0/html/123.

Using a replacement names this concept and trims down the templating code to not need to check the path prefixes.

### DIFF
--- a/_data/documentation/0_10_0.yaml
+++ b/_data/documentation/0_10_0.yaml
@@ -3,7 +3,7 @@ docs:
     description: Start here if you're experimenting with the proxy for the first time.
     tags:
       - proxy
-    path: html/proxy-quick-start/
+    path: $(DOC_BASE_FOR_VERSION)/html/proxy-quick-start/
     rank: '000'
   - title: Proxy Guide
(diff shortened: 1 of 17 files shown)


tombentley (commented):
It would be more natural to use `page.url`. I tried that before, but maybe we can make it work using {% render %}?

tombentley:
Oh, also you'll need to fix all the files in `_data/documentation`.

robobario:
> It would be more natural to use `page.url`. I tried that before, but maybe we can make it work using {% render %}?

you mean have the data declare something like:

`$(VERSION)/html/abc`

and have this interpreted as relative to the `page.url`?

robobario:
what do you think of merging `release` and `documentation` data so there's 1 data file per release?

robobario:
nah, happy to close, we can refactor if/when we add further paths

PR #157, opened 2025-07-03
Per-version quick starts

tombentley:
* Moves the quickstarts under /documentation
* Parameterises the content on the version (e.g. for the release download).
* Includes a fix for the switch to `filterDefinitions` we did in 0.10.0, and also a switch in file extension `.yml` -> `.yaml`
* No proxy quickstart before and including 0.3.0, since we didn't publish the relevant asset.
* I have not re-verified that all these quickstarts actually work as described with kroxy as it was at that point in time. No one really cares at this point. But we could fix issues if they were reported.


robobario:
Do you see this becoming another pair of guides delivered by the other repository per release?

tombentley:
In short, yes. See https://github.com/kroxylicious/kroxylicious/issues/2401 and https://github.com/kroxylicious/kroxylicious/issues/2402.

PR #159, opened 2025-07-03
Refactor release-archives page to use a custom liquid filter

robobario:
The `sort_by_semver_key` filter takes a hash and sorts it by the key, emitting an array of key-value arrays. It can take either a dotty or underscorey semver key and sort them in descending order. They are sorted lexicographically by left padding each version component with zeros, so might need some work to handle SNAPSHOT if we ever want that involved.

The filter can be configured to sort them ascending with `sort_by_semver_key: 'asc'`.


robobario:
Alternative to #151 that doesn't require us to duplicate the version or add a rank property to the data.

tombentley (approved):

release-archives/index.md
```diff
@@ -2,56 +2,26 @@
...
-  {%- else -%}
-    {%- break -%}
-  {%- endif -%}
-{%- endfor -%}
-{%- assign versionsList=versionsList | remove_first: "," | split: "," | reverse -%}
-{%- comment -%}
-endmonstrosity
-{%- endcomment -%}
+
+{%- assign sorted_releases = site.data.release | sort_by_semver_key -%}
```
robobario, line 6: will add a comment that this is custom

PR #162, opened 2025-07-10
Enable users to manually trigger a deployment from target branch

robobario:
When testing on a fork I want to be able to push to a non-main branch and deploy that, rather than force-pushing to main, which is a bad habit to get into.


SamBarker (approved):
thanks @robobario

PR #163, opened 2025-07-11
Generate redirects

SamBarker:
Generate redirects from short/stable URLs to documentation so we can update if they ever become stale.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -96,7 +96,6 @@ GEM
       google-protobuf (~> 4.31)
     sass-embedded (1.89.2-x86_64-linux-gnu)
       google-protobuf (~> 4.31)
-    semver-string (1.0.0)
     terminal-table (3.0.2)
       unicode-display_width (>= 1.1.1, < 3)
     unicode-display_width (2.6.0)
@@ -114,7 +113,6 @@ DEPENDENCIES
   jekyll-sass-converter (~> 3.1.0)
   jekyll-seo-tag (~> 2.8.0)
   jekyll-toc (~> 0.19.0)
-  semver-string (~> 1)
 
 BUNDLED WITH
    2.4.22


Gemfile.lock
```diff
@@ -94,13 +104,15 @@ GEM
...
 PLATFORMS
+  arm64-darwin-24
   x86_64-linux
 
 DEPENDENCIES
   asciidoctor-diagram (~> 2.3.2)
   jekyll (~> 4.4.1)
   jekyll-asciidoc (~> 3.0.1)
   jekyll-feed (~> 0.12)
+  jekyll-redirect-from
```
robobario, line 115: I think we're missing a corresponding `Gemfile` change like:

```
diff --git a/Gemfile b/Gemfile
index 5423258..86df08d 100644
--- a/Gemfile
+++ b/Gemfile
@@ -16,4 +16,5 @@ group :jekyll_plugins do
   gem "jekyll-sass-converter", "~> 3.1.0"
   gem "jekyll-seo-tag", "~> 2.8.0"
   gem "jekyll-toc", "~> 0.19.0"
+  gem 'jekyll-redirect-from', '~> 0.16.0'
 end
```

SamBarker: Oh yeah I was starting to think I didn't need the plugin. Just an incomplete revert.

robobario: or was the plugin an aborted experiment, I guess the page generator is a separate thing?

robobario:
PR to use the layout https://github.com/SamBarker/kroxylicious.github.io/pull/3

`fromVersion` is a bit of funny one, maybe we should have a placeholder that goes in the mapping file to use it like `${fromVersion}/html/my/doc`.

Also, I wonder if `baseUrl` should specify `kroxylicious.io` directly, it's part of the Jekyll site config, I imagine we can get at it programmatically. Though being able to redirect to other domains might be handy if we ended up with content anywhere else.

robobario:
Mostly wondering what the process would be to keep it in sync across releases. Like it the codebase refers to `kroxylicious.io/redirects/errors/${latestVersion}/specificError`, we need to ensure we create a new redirect for each version of the code we release. If a particular error is no longer relevant to the latest version, we probably want to stop publishing versioned redirects at that point.

SamBarker:
> or, introduce an optional toRelease field in the metadata and then potentially use the other data files (use them to detect all available versions) to generate a redirect for each known release that's in the range of targeted releases. (This mechanism would rely on the redirect path into the documentation being the same for each version, which I don't think we want, since we may shuffle around where the relevant docs live across versions)

I went with this option with a slight twist: `toRelease` is optional and will default to the `latestRelease` as understood by the website repo. However the slug that gets redirected will include a versions so we end up with `/redirects/erorrs/0.13.0/suspectedTls`. That way If we move the `Securing connections from clients` section in the guide we can set a `toVersion` and it will generated `/redirects/erorrs/[0.13.0..0.15.0]/suspectedTls` pointing at the current location and `/redirects/erorrs/0.16.0/suspectedTls` pointing at the new location. 

>Also, I wonder if baseUrl should specify kroxylicious.io directly, it's part of the Jekyll site config, I imagine we can get at it programmatically. Though being able to redirect to other domains might be handy if we ended up with content anywhere else.

Yeah, I'm hesitating about that. I went with configurable to support forks and redirecting to say the JDK docs, but I do wonder about the security implications of allowing arbitrary re-directs. Given its a static site and version controlled we are probably ok. I was mussing about an allow list of domains, but that would be updated by the same mechanisms as the redirect plugin so doesn't really add any security. 
Maybe it should be optional and default to the jekyll site?

SamBarker:
This can be seen working on `[redirect/errors/0.13.0/suspectedTls](https://sambarker.github.io/kroxylicious.github.io/redirect/errors/0.13.0/suspectedTls)`

robobario:
> Thinking a bit about [kroxylicious/kroxylicious#2422 (comment)](https://github.com/kroxylicious/kroxylicious/pull/2422#discussion_r2204956525) should we also generate `latestRelease+1-SNAPSHOT` ? So that dev links work? Should the link generator only deal in released versions?!

I don't think so. I think the prod website should deal in the released versions only.

For local dev we could make it convenient with some scripting to do something like:

1. build `target/web` in `kroxylicious`
2. cp -r `target/web` into the website repo
3. generate a faked out release YAML for the snapshot version
4. build and run it (what the `run.sh` does currently)

Then we could build and run it. Maybe we would also need to override latestRelease to the snapshot version? I'm guessing there's a bunch of wrinkles hidden in there.

I'm also wondering if we want to do that exact same process in the github pages we deploy to `kroxylicious.io/kroxylicious` to have a continuously deployed development version of the website with the latest documentation off main installed into it.

robobario:
the release flow is also a bit interesting, it makes things a little less automatable.

If you want to change to a new location for the new version, or add in more links for the new version, how does that look at release time?

1. merge in `target/web` (create a PR with automation)
2. update the mapping files with new/updated links (manual step?)
3. merge PR

Maybe the links for a version should be delivered from the `kroxylicious` repo, if we say restrict these links to: "version specific documentation links" then it makes sense for them to be sourced from the other repo along with that version of the documentation.

Happy to merge and iterate on it though, the generation code is going to be used in whatever scheme.

robobario (approved):
The redirect page and generation code looks good, but I'm wondering if we should keep iterating so that the version-specific documentation links would be delivered as part of the files emitted by the `kroxylicious` build. 

That would make a logical bundle, new links could arrive in the same bundle as the new documentation. On the kroxy side you could have a nice PR containing new docs, a new redirect configuration, and the code referencing the redirect-link.

robobario:
> > also makes me wonder about the state of testing on this repo, without going too crazy on it this particular bit of logic feels nice to have tests on (and the output HTML is simple enough that we aren't looking at anything hellish to grok the content), maybe we could include some CI that runs the production docker build to check it's working and also could inspect the generated redirect files to test they point where we want
> > edit: non-blocking, I'm keen to have a go at adding the CI
> 
> `/me mumbles: something something rspec` thus completely exhausting his knowledge of ruby testing

meanwhile, you styled the page up to look fancy :grin: so it's not just a tiny snip of html to compare, oh well

SamBarker:
> The redirect page and generation code looks good, but I'm wondering if we should keep iterating so that the version-specific documentation links would be delivered as part of the files emitted by the `kroxylicious` build. 
> 
> 
> 
> That would make a logical bundle, new links could arrive in the same bundle as the new documentation. On the kroxy side you could have a nice PR containing new docs, a new redirect configuration, and the code referencing the redirect-link.

I had envisioned a pr flow to add new slugs, hence the from & to version business. I liked the fact that a human could spot a slug being deleted (shouldn't happen after the fromVersion is released).

 Maybe replacing the redirect file as part of the push from the dev docs would work. Maybe I should revisit the idea of one file per version but I'd like to do that in a new pr.

robobario:
maybe it would be cool to have the source of truth be in kroxy code, like have an enum which has the relative links in them like `CLIENT_TLS("/path/to/tls#section")` and generate a per-version links config file from the enum. Would help keen them in sync.

_plugins/redirector.rb
```diff
@@ -0,0 +1,135 @@
...
+    attr_reader :patch
+
+    def self.parse(version_string)
+      if version_string.is_a? Version
+        return version_string
+      end
+      if version_string.include? '_'
+        parts = version_string.split('_').map { |x| x.to_i }
+      elsif version_string.include? '.'
+        parts = version_string.split('.').map { |x| x.to_i }
```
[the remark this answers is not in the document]
SamBarker: yeah. I think your right we don't need to handle it however.

PR #166, opened 2025-07-15
Add Development Mode warning.

robobario:
Adds a transparent warning to the top right of the page when in development environment. This is the default jekyll environment so in the action I'm setting it to production for the real deal, though nothing acts on the "production" value currently. The warning does not interfere with the other content and clicks "pass through" the div, so if it happens to overlap the nav items they are clickable.

We are considering deploying a development verstion of the website continuously to GitHub pages somewhere, so we need a way to disambiguate from the main page in case someone stumbles on it. It will also make it immediately clear when developing locally that we are looking at a dev deployment.

Screenshot From 2025-07-16 09-14-34


.github/workflows/jekyll-gh-pages.yml
```diff
@@ -70,7 +70,7 @@ jobs:
             -u "$(id -u):$(id -g)" \
             -v "$(pwd):/site" \
             localhost:5000/kroxy-jekyll:latest \
-            bash -c 'eval "$(rbenv init -)" && cp -r /css/_sass/bootstrap /site/_sass/ && bundle exec jekyll build --config=_config.yml,_config-overrides.yml'
+            bash -c 'eval "$(rbenv init -)" && cp -r /css/_sass/bootstrap /site/_sass/ && JEKYLL_ENV=production bundle exec jekyll build --config=_config.yml,_config-overrides.yml'
```
robobario, line 73: will test this on my fork

robobario: LGTM, deployed up with no warning

PR #167, opened 2025-07-15
Add CI to check that we can build the production site

robobario:
This adds a separate workflow for PR CI, just checking we can build the site in production mode. We might need an accompanying change in the settings to require the checks to succeed before users can merge PRs.

The workflow is similar to the automation that deploys the website to GitHub Pages, but I made it a separate workflow as I'm envisaging adding specializations here. Like if there is any linting or simple tests we want to execute against the built site HTML we could do them in this workflow without riddling the prod deploy workflow with conditionals.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -1,0 +1,41 @@
+name: Test PR
+
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


SamBarker (approved):
Probably worth addressing the GHA warnings, but I think this is a very useful addition.

PR #171, opened 2025-07-20
Tolerate snapshot release _data

robobario:
I want to install a mocked out release YAML and overwrite the currentVersion in the snapshot website so that the new release appears in the release-archive and is the first release you see at http://kroxylicious.io/kroxylicious/documentation/


k-wall (approved):
looks reasonable to me.

PR #175, opened 2025-07-29
Fix links for 0.13.0

robobario:
1. Some links and text refers to 0.13 where it should be 0.13.0
2. Some links referred to main where they should point at a version specific location on github, tied to the release tag.

Partially resolves https://github.com/kroxylicious/kroxylicious/issues/2478


tombentley (commented):
This is the post-asciidoctor content, so did we/are we going to fix the source adoc?

SamBarker (approved):
LGTM

PR #177, opened 2025-07-30
Add build script

robobario:
Add a build script so that we can build the same way locally and in CI

I'm experimenting with some automation for checking that all the links within the generated HTML are valid, having a way to produce the production build output locally is helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,13 +23,27 @@ We don't use the Bootstrap 5 ruby gem here, as it runs on a
 
 ### Running in a Container
 
+When developing website content (editing markdown/asciidoc, modifying ruby extensions) it is useful to deploy a
+local testing server that presents the website, with hot-reloading so that source changes are reflected quickly
+in the built HTML.
+
 To build and serve the website from a container you can run `./run.sh`. It will be deployed on http://localhost:4000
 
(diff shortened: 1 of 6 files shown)


tombentley:
> Add a build script so that we can build the same way locally and in CI

I think you should explain in the README how and why people would run this locally

robobario:
done

robobario:
I've tested the github automation on my fork

main run: https://github.com/robobario/kroxylicious.github.io/actions/runs/16611488912
PR test: https://github.com/robobario/kroxylicious.github.io/actions/runs/16611509577

README.md
```diff
@@ -47,6 +47,13 @@ To enable pages on your fork:
 8. Push changes to any branch of your fork and then trigger a manual run of `https://github.com/${yourname}/kroxylicious.github.io/actions/workflows/jekyll-gh-pages.yml`,
    supplying the branch you want to checkout and deploy as a parameter. 
 
+### Building Production HTML
+
+Sometimes it can be useful to build the same static output that is deployed to GitHub Pages. For example, you may need to test some Jekyll
+changes produce the URLs you expect in production.
```
tombentley, line 53: I think you should spell out a little more clearly what the difference between `build.sh` and `run.sh` is, in use case terms. And probably move this section before the section about running on a fork (which seems to me like a more advanced thing to want to do)

PR #179, opened 2025-08-04
Fail PR build if any image SHA placeholders are found

robobario:
The intent is to prevent documentation PRs from merging until we manually update the PR with the SHAs.

See matching PR https://github.com/kroxylicious/kroxylicious/pull/2529

The image shas are only available after the image build, which happens in response to the release tag being pushed up. So they are not available at the point we are currently creating the documentation PR.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -38,3 +38,19 @@ jobs:
         env:
           CONTAINER_ENGINE: docker
           BUILD_IMAGE_SPEC: localhost:5000/kroxy-jekyll:latest
+
+      # Currently the image shas are not available when we create the release, they need to be manually
+      # updated in the release PR before merge.
+      - name: Require release manifests to contain non-placeholder container image SHAs
+        run: |
+          PLACEHOLDER="REPLACE_WITH_SHA_AFTER_IMAGE_RELEASE"
+          # || true prevents the script failing when 0 matches are found
+          matching_files=$(grep -r -l "${PLACEHOLDER}" _data/release/ || true)
+          if [[ -n "${matching_files}" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-build.yml
```diff
@@ -38,3 +38,19 @@ jobs:
         env:
           CONTAINER_ENGINE: docker
           BUILD_IMAGE_SPEC: localhost:5000/kroxy-jekyll:latest
+
+      # Currently the image shas are not available when we create the release, they need to be manually
+      # updated in the release PR before merge.
+      - name: Require release manifests to contain non-placeholder container image SHAs
+        run: |
+          PLACEHOLDER="REPLACE_WITH_SHA_AFTER_IMAGE_RELEASE"
+          # || true prevents the script failing when 0 matches are found
```
tombentley, line 47: Is the `|| true` part commented out code?

.github/workflows/pr-build.yml
```diff
@@ -38,3 +38,19 @@ jobs:
...
+          matching_files=$(grep -r -l "${PLACEHOLDER}" _data/release/ || true)
```
tombentley, line 48: I guess for should add `-F`, because you've not said that `PLACEHOLDER` is intended to be a regex.

tombentley (approved):
A couple of nits, otherwise LGTM. Thanks.

.github/workflows/pr-build.yml
tombentley: Some how I didn't connect it up with the following line. I would write this as

```
# The `|| true` below prevents the script failing when 0 matches are found
```

But I don't think it's worth a PR to change it.

PR #183, opened 2025-08-05
Use a boring old table for documentation links

robobario:
Cards were getting big v0.14.0 is going to introduce even more. Even nicer might be a tree like arrangement with quickstarts/guides/other grouped together.

Before:
Screenshot From 2025-08-05 17-24-40

After:
Screenshot From 2025-08-05 17-23-37

### DIFF
--- a/_layouts/released-documentation.html
+++ b/_layouts/released-documentation.html
@@ -35,7 +35,12 @@ layout: default
     </div>
   </div>
   <div class="col-lg-6" role="main">
-    <div class="row row-cols-1 row-cols-md-2 g-4">
+    <div class="row">
+    <table class="table">
+      <thead>
+      <tr><th class="col-lg-2">Document</th><th class="col-lg-4">Description</th></tr>
+      </thead>
+      <tbody>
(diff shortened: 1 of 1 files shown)


tombentley:
I agree that the number of guides is becoming a bit of a problem. But I'm not sure switching to a table is this best way to tackle this. It make the page less visually appealing. 
* I'd like to move more in the direction of what quarkus has ([see here](https://quarkus.io/guides/#)), making use of the documentation metadata (e.g. `tags`), so visitors can easily select content based on what they're trying to achieve. I don't think we need their text search functionality yet though.
* I'd also like to make these page a bit more visually interesting with icons, e.g Duke for stuff aimed at Java programmers, the Kube ships wheel icon for kube stuff, and maybe the Kroxylicious crocodile for straight proxy stuff? We'd need to confirm that using these icons like this is OK per the wishes of the the trademark owners. I _think_ it is, but let's point to something specific in any PR which does this so we have a record that we did it.

robobario:
fair enough, maybe they could just be compacted slightly. Instead of text search maybe it would be nice to be able to filter them by purpose like you've said, like if the same icons were up at the top or side bar, and clicking them reduced it down to only developer or k8s cards

PR #188, opened 2025-08-14
Improve guide style

robobario:
Aligns it more with the standalone asciidoc style, which has much nicer margins/padding and a single pixel top border for main sections, more padding on the codeblocks.

Adds a class to the top div surrounding the content so we can specifically target CSS to the guide content.

## Before
Screenshot From 2025-08-14 16-44-03

## After
Screenshot From 2025-08-14 16-43-29


robobario:
The proxy quick start also includes some horizontal rules which I think we should drop in favour of styling the sections with css.

PR #192, opened 2025-08-20
Add copy to clipboard button to code blocks within documentation

k-wall:
Our documentation sometimes contains code and config blocks.  This is a especially true in the quickstarts where we present a list of instructions that the user is expected to follow.  We want to make it easy for the users to copy the content from the documentation and paste it in their terminal or IDE.

This change add this function to the documentation on the website.

Approach based on a this SO article:

https://stackoverflow.com/questions/48078199/jekyll-code-snippet-copy-to-clipboard-button

Screenshot 2025-08-21 at 12 48 45

### DIFF
--- a/_includes/asciidoctor-scripts/clipboard.html
+++ b/_includes/asciidoctor-scripts/clipboard.html
@@ -6,16 +6,16 @@
     document.querySelectorAll("code").forEach((codeElem) => {
         if (codeElem.textContent && codeElem.closest('.listingblock')) {
             let selectableCode = codeElem.textContent;
-            if (codeElem.getAttribute("data-lang") === "terminal") {
+            if (codeElem.dataset.lang === "terminal") {
                 let codeLines = codeElem.textContent.split(/\n/);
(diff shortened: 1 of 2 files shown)


_layouts/guide.html
```diff
@@ -35,5 +43,31 @@ <h1 class="card-title fs-3">
...
+
+allCodeBlocksElements.each(function(i) {
+
+if ($(this).parents('.listingblock').length) {
+
+  var code = $(this).text().split(/\n/);
+  for (var idx = 0; idx < code.length; idx++) { code[idx] = code[idx].replace("$ ", "") }
+
+  //trigger
+  var clipButton = '<button class="btn" data-clipboard-text="' + code.join("\n") + '"><img src="https://clipboardjs.com/assets/images/clippy.svg" width="13" alt="Copy to clipboard"></button>';
```
k-wall, line 59: Is clippy's license okay?
Make resource part of the site

k-wall: MIT licensed.

PR #213, opened 2026-01-05
Fix #238: Publish Javadoc on Website - PR 1 of 2

### LINKED ISSUE #238: Javadoc links broken
**Suggestion / Problem**
Some links in our javadoc refer to kroxylicious.io/kroxylicious/* urls when linking to documentation from another module.

For example, in the [ApiVersionsFilter](https://javadoc.io/doc/io.kroxylicious/kroxylicious/latest/io/kroxylicious/proxy/internal/filter/ApiVersionsFilter.html) javadoc it links to https://kroxylicious.io/kroxylicious/api/kroxylicious-filter-api/apidocs/io/kroxylicious/proxy/filter/ApiVersionsResponseFilter.html which is a 404

gracegrimwood:
This PR is the website portion of the work to fix https://github.com/kroxylicious/kroxylicious/issues/238 (Javadoc redirect links broken). This sets up the scaffolding on the website side for getting the Javadoc published — there will be a second PR **(edit: [this one](https://github.com/kroxylicious/kroxylicious/pull/3096))** on the Kroxylicious repo side to aggregate that Javadoc and push it into this repo during releases alongside the current Asciidoc push.

Things not included in this PR that could or possibly should be: _(I ran out of time to finish them, because I spent far too long arguing with CSS conflicts)_

- Haven't got a good way of dealing with multiple versions of Javadoc. Not a problem if we only care about the latest, might become a problem if we want to have a historical record on the site.
- The current mechanism used to get the Javadoc to render nicely and also get the Kroxylicious website header over top of it is an iframe, which is a bit of a risky move because it's been talked about being deprecated for decades. Alas, the other option was rewriting either the Javadoc styles or the website styles from scratch, and so using iframes seemed like the easier option.
- The way the files get put into this repo so that they play nice with Jekyll is kind of gross (the script for this will be in the other PR). It copies all the files over, and then for every HTML file it generates a _second_ blank HTML file for us to put the iframe into and use to render the canonical links. It's kinda grotty, but it works.

**NB:** I've left IBM now, but I still have a laptop and slack and email, so just let me know if I need to put a bit more work into this. I just may not be as quick to respond as I used to be.


tombentley:
We solved this in a different way in the end.

PR #248, opened 2026-04-09
Tweak community call recordings link

tombentley:
The recording are all in the "live" section on you tube, so it would be better to link there directly, than just to the our channel, which includes all the other content too.


SamBarker (approved):
LGTM

PR #250, opened 2026-04-22
Remove github discussions

robobario:
Remove github discussions as we've deactivated it for being an relatively dead comms channel


tombentley (approved):
Thanks @robobario

PR #251, opened 2026-04-29
feat: add release schedule page

robobario:
Adds a release schedule page to provide visibility into planned and released versions of Kroxylicious. This helps users and contributors track upcoming releases and their target dates.

The page includes milestones linked to GitHub for detailed tracking.

Adds a Footer link

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>

image

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -52,6 +52,11 @@
                         </a>
                     </li>
                     <li class="mb-2">
+                        <a class="text-body link-underline-dark link-underline-opacity-0 link-underline-opacity-10-hover" href="{{ '/release-schedule/' | absolute_url }}">
+                            Release Schedule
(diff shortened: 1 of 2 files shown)


SamBarker (approved):
LGTM 

would be great to tie this into project/milestone due dates in GitHub but lets start with something simple.

robobario:
yeap, could also imagine having some kind of integration with the release metadata, so we could populate an entry in the table for each metadata file in `_data/release` with status 'released' and link to something relevant.

Hmm, I wonder if the planned releases could have a metadata file each that are ignored by the other bits of the website. Then at release time it's overwritten.

robobario:
will wait til we've got some agreement on the dates on the dev-list

PR #252, opened 2026-05-04
Update snapshot docs script

robobario:
We need to build dependencies of kroxylicious-docs too as it is now responsible for copying over javadoc from API modules.


robobario:
closing, this change was included in #256

PR #258, opened 2026-05-18
feat: automate release status in schedule

robobario:
Convert release schedule from static HTML to data-driven Liquid templating. Status now automatically changes from "Planned" to "Released" when a release YAML file appears in _data/release/, eliminating manual updates when releases are published.

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>

Screenshot From 2026-05-18 16-47-44

### DIFF
--- a/_data/release-schedule.yml
+++ b/_data/release-schedule.yml
@@ -1,0 +1,7 @@
+releases:
+  - version: 0.21.0
+    plannedDate: May 15, 2026
+    milestoneUrl: https://github.com/kroxylicious/kroxylicious/milestone/27
+  - version: 0.22.0
+    plannedDate: July 3, 2026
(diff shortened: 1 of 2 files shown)


SamBarker (approved):
Getting there. I still wonder if we can drive it directly from the github milestone dates?

robobario:
> Getting there. I still wonder if we can drive it directly from the github milestone dates?

Will look into it, it would be cool if creating/editing a milestone would cause a PR to be created here.

robobario:
Also, I currently don't think there's a good way to figure out the canonical "release date", the per release files don't contain that currently. It's one that could also get out of sync between creating the release PRs and hitting merge.

Thinking it might be better if the Column was called "Release Date" and already released ones display their release date, and planned ones put their intended date from the schedule.

PR #261, opened 2026-06-03
docs(blog): add benchmarking under-the-hood post with redirect infrastructure

SamBarker:
## Summary

- Adds the companion engineering deep-dive post: _Does my proxy look big in this cluster? — Under the hood_
- Replaces placeholder flamegraph files with correct-cluster data and recomputes CPU breakdown tables
- Extends the redirector plugin with `absoluteTarget` support for version-free external redirects, with a `subgroup` field for nested URL paths
- Adds `blog.yaml` redirect mapping linking to raw benchmark data (OMB results, metrics, JFR recordings) on Google Drive via `/redirect/blog/benchmarking-the-proxy-under-the-hood/benchmark-data`
- Fixes Post 1 doc link (0.20.0 → 0.21.0) and publication date

## Test plan

- [ ] `./run.sh` builds without errors
- [ ] `http://127.0.0.1:4000/redirect/blog/benchmarking-the-proxy-under-the-hood/benchmark-data` redirects to Google Drive folder
- [ ] Existing redirect URLs (`/redirect/errors/`, `/redirect/qr-codes/`) still work
- [ ] Both posts render correctly at their publication URLs

🤖 Generated with [Claude Code](https://claude.com/claude-code)


_posts/2026-06-03-benchmarking-the-proxy-under-the-hood.md
```diff
@@ -0,0 +1,307 @@
...
+
+At 164 MB/s across 10 partitions, each partition carries 16 MB/s — within the budget of a single Netty thread. The elevated E2E latency reflects the proxy running near but not at its ceiling. The 2-core saturation point sits above 164 MB/s; the model is consistent.
+
+The full picture — coefficient measured across all three core counts and three topic counts:
+
+| | 1-core (1000m) | 2-core (2000m) | 4-core (4000m) |
+|---|---|---|---|
+| **1 topic** | 15.7 mc/MB/s | 24.3 mc/MB/s | 30.5 mc/MB/s† |
+| **10 topics** | 10.0 mc/MB/s | 19.9 mc/MB/s | 25.1 mc/MB/s |
+| **100 topics** | 4.3 mc/MB/s | 6.8 mc/MB/s | 8.0 mc/MB/s |
```
robobario, line 185: claude was complaining that this table is a bit hard to square with the linear scaling claim, saying a constant coefficient would be expected in that case. maybe just needs some further explanation?

robobario (approved):
LGTM, claude suggesting the triple question-mark is a bit much in the title, too social-media-ey. Also that it's quite aside heavy so maybe one or two of them could be dropped `(WHAT BROKE NOW??) is the most expendable`

_posts/2026-06-03-benchmarking-the-proxy-under-the-hood.md
SamBarker: Good catch. After digging into the JFR data from the coefficient sweep, the short answer is: the rising coefficient is real, not noise, but the sweep methodology conflates two effects (more connections *and* more throughput at each step), so we can't cleanly attribute it yet.

Preliminary regression on the non-saturated probes suggests the per-byte encryption cost is roughly constant across core counts (~38 mc/MB/s for 1-core and 2-core), but the baseline JVM/Netty thread overhead scales with core count — each additional Netty event loop thread adds a fixed per-thread cost regardless of how many bytes it's processing. The proper experiment to confirm this is a rate sweep at fixed connection count, which hasn't been run yet.

I've added a footnote to the table acknowledging this and pointing readers to the ceiling table as the cleaner test of linear scaling. The clean sweep is on the todo list for a follow-up post.

PR #265, opened 2026-07-01
Move 0.23.0 forward

robobario:


robobario:
Keith was happy to move the date forward in call, so think it just needs a thumb up from @SamBarker and we can merge it.

SamBarker (approved):
LGTM

per call and mailing list

